PR #23, opened 2020-12-23
Implementation of Landsat STAC handling

alexgleith:
* Half-finished effort at handling MTL files, but aborted because they don't have all the information needed
* Mostly complete implementation of STAC 0.7.0 to STAC 1.0.0.beta.2
* Functions are kind of tested, could do with some eyeballs to see if I'm doing anything foolish
* CLI needs thought, to work out what's useful


stactools_landsat/setup.py
```diff
@@ -4,42 +4,44 @@
...
+    author="stac-utils",
+    author_email="stac@radiant.earth",
+    url="https://github.com/stac-utils/stactools.git",
+    packages=find_namespace_packages(),
+    py_modules=[splitext(basename(path))[0] for path in glob("stactools/*.py")],
+    include_package_data=False,
+    install_requires=install_requires,
+    license="Apache Software License 2.0",
+    keywords=["stactools", "psytac", "imagery", "raster", "catalog", "STAC"],
+)
```
lossyrob, line 47: Looks like your editor is formatting code in a way that's different than what the project expects - you can use `scripts/format` to format the code with yapf; that will fix the linting stage of the CI. Running `scripts/test` will fail if there's any linting failures from yapf or flake8 (or if the spellcheck or unit tests fail).

stactools_landsat/stactools/landsat/commands.py
```diff
@@ -1,16 +1,35 @@
...
-    def convert_command(mlt, dst):
-        convert_mtl_to_item(mlt, dst)
+        metadata = None
+        with open(source) as f:
+            metadata = json.load(f)
+
+        result = transform(metadata)
+
+        with open(dst, "w") as outfile:
+            json.dump(result, outfile)
```
lossyrob, line 33: Would be better for the library commands to return the `pystac.Item` directly, and then to use `item.save_object` directly a la [this example](https://github.com/stac-utils/stactools/blob/2c6a1eab7acfb50d72c735ce9508012efc7d75d3/stactools_aster/stactools/aster/commands.py#L49-L57). This will use PySTAC's STAC_IO, which is setup to utilize fsspec and abstract over different file systems.

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,124 @@
+import datetime
+
+import dateutil
+import rasterio
+from pystac import Item, Link
+from rasterio import RasterioIOError
+from shapely.geometry import box, mapping, shape
+
+
+def _parse_date(in_date: str) -> datetime.datetime:
```
lossyrob, line 10: Use [pystac.utils.datetime_to_str](https://github.com/stac-utils/pystac/blob/develop/pystac/utils.py#L144) (has more generality and is better tested)

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    dt = dateutil.parser.parse(in_date)
+    return dt.replace(tzinfo=datetime.timezone.utc)
+
+
+def transform_mtl_to_item(metadata: dict) -> dict:
+    """
+    Handle USGS MTL as a dict and return a STAC dict.
+
+    This isn't implemented, because all the info we need
```
lossyrob, line 22: Confused by this comment - do you mean it's not used?

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    item.common_metadata.instruments = [
+        i.lower() for i in image["SENSOR_ID"].split("_")
+    ]
+
+    # eo
+    item.ext.enable("eo")
+    item.ext.enable("view")
+    item.ext.enable("projection")
+
+    item.validate()
```
lossyrob, line 60: The extensions are enabled, but the extension properties are not filled out - is that a TODO?

stactools_landsat/tests/test_landsat.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_transform_dynamic_stac(landsat_stac_api_one):
+    result = stac_api_to_stac(landsat_stac_api_one)
+
+    item = Item.from_dict(result)
+    item.validate()
+
+
+@pytest.fixture
```
lossyrob, line 49: Heads up for the post-Draft PR, we use `unittest` and not `pytest` in the CI, so better to structure this like the other unit tests in the project as `unittest.TestCase`s.

lossyrob (commented):
Thanks for this PR! This is looking like a great start. Added some comments inline. Let me know if you want to strategize around moving this from a Draft to a ready PR; I'll hold off more comments until this is further along.

stactools_landsat/setup.py
alexgleith: Ok, got it.

stactools_landsat/stactools/landsat/utils.py
alexgleith: I see. I'll look into it, but I think I see what you're saying. Parse the date here using the generic tool, and make it a string using the `datetime_to_str` function.

stactools_landsat/stactools/landsat/utils.py
alexgleith: This was only started and not completed, because the MTL is missing a lot of information...

I don't think using the MTL is a good idea at all!

stactools_landsat/stactools/landsat/utils.py
alexgleith: Yep, not complete. Probably should delete the MTL function.

stactools_landsat/tests/test_landsat.py
alexgleith: Ah, I see. I'll go hunting for other examples and re-implement in unittest.

stactools_landsat/stactools/landsat/utils.py
lossyrob: Ack, nope. Sorry, got my wires crossed, through this was going datetime -> str, it's going str -> datetime.

[pystac.utils.str_to_datetime](https://github.com/stac-utils/pystac/blob/develop/pystac/utils.py#L164-L165) just uses the dateutil parse, and this is setting the TZ, so disregard my comment!

stactools_landsat/stactools/landsat/utils.py
lossyrob: That makes sense, if it's not useful to you, we'll wait until someone needs it to implement :-)

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def transform_mtl_to_stac(metadata: dict) -> Item:
+    """
+    Handle USGS MTL as a dict and return a STAC item.
+
+    NOT IMPLEMENTED
+
+    Issues include:
+        - There's no reference to UTM Zone or any other CRS info in the MTL
```
kylebarron, line 25: Is the `PROJECTION_ATTRIBUTES` section not enough? It looks like maybe it doesn't say whether the utm zone is north or south... but I'd guess you could deduce it.

```bash
> aws s3 cp s3://usgs-landsat/collection02/level-2/standard/oli-tirs/2020/035/033/LC08_L2SP_035033_20200725_20200908_02_T2/LC08_L2SP_035033_20200725_20200908_02_T2_MTL.json - --request-payer | jq '.LANDSAT_METADATA_FILE.PROJECTION_ATTRIBUTES'
```

```json
{
  "MAP_PROJECTION": "UTM",
  "DATUM": "WGS84",
  "ELLIPSOID": "WGS84",
  "UTM_ZONE": "13",
  "GRID_CELL_SIZE_REFLECTIVE": "30.00",
  "GRID_CELL_SIZE_THERMAL": "30.00",
  "REFLECTIVE_LINES": "8001",
  "REFLECTIVE_SAMPLES": "7871",
  "THERMAL_LINES": "8001",
  "THERMAL_SAMPLES": "7871",
  "ORIENTATION": "NORTH_UP",
  "CORNER_UL_LAT_PRODUCT": "39.93756",
  "CORNER_UL_LON_PRODUCT": "-109.18018",
  "CORNER_UR_LAT_PRODUCT": "40.00431",
  "CORNER_UR_LON_PRODUCT": "-106.41877",
  "CORNER_LL_LAT_PRODUCT": "37.78038",
  "CORNER_LL_LON_PRODUCT": "-109.05554",
  "CORNER_LR_LAT_PRODUCT": "37.84221",
  "CORNER_LR_LON_PRODUCT": "-106.37634",
  "CORNER_UL_PROJECTION_X_PRODUCT": "142800.000",
  "CORNER_UL_PROJECTION_Y_PRODUCT": "4429200.000",
  "CORNER_UR_PROJECTION_X_PRODUCT": "378900.000",
  "CORNER_UR_PROJECTION_Y_PRODUCT": "4429200.000",
  "CORNER_LL_PROJECTION_X_PRODUCT": "142800.000",
  "CORNER_LL_PROJECTION_Y_PRODUCT": "4189200.000",
  "CORNER_LR_PROJECTION_X_PRODUCT": "378900.000",
  "CORNER_LR_PROJECTION_Y_PRODUCT": "4189200.000"
}
```

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,141 @@
...
+    # Handle view extension
+    item.ext.enable("view")
+    item.ext.view.off_nadir = item.properties["eo:off_nadir"]
+    del item.properties["eo:off_nadir"]
+
+    if enable_proj:
+        try:
+            # If we can load the blue band, use it to add proj information
+            blue_asset = item.assets["SR_B2.TIF"]
+            blue = rasterio.open(blue_asset.href)
```
kylebarron, line 104: If you use the MTL file instead, I don't think you need to load the asset.. Specifically, the bounding box in the local crs is included, so you should be able to infer the transform

This should be all you need?
```json
{
  "MAP_PROJECTION": "UTM",
  "UTM_ZONE": "13",
  "GRID_CELL_SIZE_REFLECTIVE": "30.00",
  "REFLECTIVE_LINES": "8001",
  "REFLECTIVE_SAMPLES": "7871",
  "CORNER_UR_PROJECTION_X_PRODUCT": "378900.000",
  "CORNER_UR_PROJECTION_Y_PRODUCT": "4429200.000",
  "CORNER_LL_PROJECTION_X_PRODUCT": "142800.000",
  "CORNER_LL_PROJECTION_Y_PRODUCT": "4189200.000",
}
```

stactools_landsat/stactools/landsat/commands.py
```diff
@@ -1,16 +1,45 @@
...
+        item = None
+        if mtl:
+            # Transform not implemented, so tell folks
+            print("MTL transform not yet implemented.")
+            sys.exit(1)
+            with open(mtl) as f:
+                item = transform_mtl_to_stac(json.load(f))
+        elif stac:
+            with open(stac) as f:
+                in_item = Item.from_dict(json.load(f))
```
lossyrob, line 37: Prefer `pystac.read_file`. You can check that the read-in stac object is an Item by checking

`in_item.STAC_OBJECT_TYPE == pystac.STACObjectType.ITEM`

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,141 @@
...
+            item.ext.enable("projection")
+            item.ext.projection.epsg = crs
+
+            new_assets = {}
+
+            for name, asset in item.assets.items():
+                if asset.media_type == "image/vnd.stac.geotiff; cloud-optimized=true":
+                    item.ext.projection.set_transform(transform, asset=asset)
+                    item.ext.projection.set_shape(shape, asset=asset)
+                    asset.media_type = "image/tiff; application=geotiff; profile=cloud-optimized"
```
lossyrob, line 119: prefer `pystac.MediaType.COG`

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+    return item
+
+
+def transform_stac_to_stac(item: Item,
+                           enable_proj: bool = True,
+                           self_link: str = None,
+                           source_link: str = None) -> Item:
+    """
+    Handle a 0.7.0 item and convert it to a 1.0.0.beta2 item.
```
lossyrob, line 73: This could use some explanation about what the expected input is...are these the STAC items published as part of the Collection 2 data?

tests/data-files/landsat/readme.md
```diff
@@ -0,0 +1,9 @@
+# Notes on data in this folder
+
+Files sourced from the landsatlook sat-api implementation, accessible here:
+`https://landsatlook.usgs.gov/sat-api/`
+
+Downloaded metadata from this scene:
+`https://landsatlook.usgs.gov/sat-api/collections/landsat-c2l2-sr/items/LC08_L2SR_081119_20200101_20200823_02_T2`
+
+Run `make download` to get these files. Run `make clean` to delete them.
```
lossyrob, line 9: Seems like these files are checked into the repo with this PR, is the makefile needed? If it's not, probably best to remove - I think the references to the download locations is enough

lossyrob (changes requested):
This is pretty close, comments inline. I think making a command that's explicitly for migrating the collection 2 published STAC items will make things more clear.

Not sure what that command should be named, but you may want to move the command inside a landsat-specific subcommand, so that you don't have to prefix with `landsat-`. E.g. https://github.com/stac-utils/stactools/blob/master/stactools_aster/stactools/aster/commands.py#L14-L24 - this would allow the command to be `stac landsast covert-c2-stac` or similar.

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -0,0 +1,141 @@
...
+    projection = LANDSAT_METADATA["PROJECTION_ATTRIBUTES"]
+    image = LANDSAT_METADATA["IMAGE_ATTRIBUTES"]
+    proessing_record = LANDSAT_METADATA["LEVEL2_PROCESSING_RECORD"]
+
+    scene_id = product["LANDSAT_PRODUCT_ID"]
+
+    xmin, xmax = float(projection["CORNER_LL_LON_PRODUCT"]), float(
+        projection["CORNER_UR_LON_PRODUCT"])
+    ymin, ymax = float(projection["CORNER_LL_LAT_PRODUCT"]), float(
+        projection["CORNER_UR_LAT_PRODUCT"])
```
kylebarron, line 39: I don't think this is correct for lon/lat coordinates. Because the image is tilted, isn't `ymin` in the lower right corner?

stactools_landsat/stactools/landsat/utils.py
kylebarron: Yeah Collection 2 includes a STAC collection but it's 0.7.0

stactools_landsat/stactools/landsat/utils.py
matthewhanson: Yes, looks like the corner projection coordinates and the grid cell size has everything to construct the affine. 
Will also need the EPSG code which can be derived (see https://github.com/sat-utils/sat-stac-landsat/blob/master/satstac/landsat/main.py#L196).

The only question I had is for other bands that are not at the nominal cell size. Thermal bands are distributed as upsampled versions to 30m, but the pan band (which is only available in the Level-1 products) is 15m. I think those projection coordinates are still valid, but may need to be shifted. Quick look shows the pan band as 2*x+1 rows/cols as opposed to the nominal 30m bands. So those projection coordinates may need to be shifted half a pixel, but I have not looked into this closely.

stactools_landsat/stactools/landsat/utils.py
matthewhanson: These coordinates actually are the bounding box coordinates, not the data footprint so the rotation doesn't factor in.

matthewhanson:
@alexgleith @lossyrob @kylebarron 
finally got around to looking through this, this is great. While the official STAC 1.0.0 metadata will be provided in the coming months, this will help that team in generating them from the existing 0.7.0 records.

@alexgleith mentioned you couldn't use the MTL files directly because it didn't have all the info. I believe you meant the absence of the data geometry (did you find any other fields that couldn't be generated?). The issue is that the data geometry isn't generated with the products themselves, but rather is determined when creating the STAC 0.7.0 metadata. I believe using a tracing algorithm. 

If we still wanted to have the ability to create STAC from the MTL files I think there are 2 options:
- use the ANG file to calculate geometry..but I'm not sure if the info required is in the Collection 2 ANG files: [implementation](https://github.com/sat-utils/sat-stac-landsat/blob/master/satstac/landsat/main.py#L108)
- use the fixed WRS-2 tiles to look up the path/row and get the geometry: [Implementation](https://github.com/sat-utils/sat-stac-landsat/blob/master/satstac/landsat/main.py#L28) but uses a [15MB look-up-table](https://github.com/sat-utils/sat-stac-landsat/blob/master/satstac/landsat/pr2coords.json)

See the README for an explanation: https://github.com/sat-utils/sat-stac-landsat/blob/master/README.md

kylebarron:
> I believe you meant the absence of the data geometry (did you find any other fields that couldn't be generated?). The issue is that the data geometry isn't generated with the products themselves, but rather is determined when creating the STAC 0.7.0 metadata.

I wouldn't call myself an expert on this, but this seems incorrect to me. I've looked at a handful of MTL files across collection 1 and 2, and they all seem to have accurate data geometries contained in the MTL file.

For each of the following examples, I look at the MTL file and a TIF using rasterio. The MTL projected data geometry lines up exactly with that `rasterio` indicates as the bounds of the image, with a half-pixel (i.e. 15-meter) buffer around the MTL's bounds. I gather that the MTL bounds are referring to the _pixel centroids_ whereas rasterio tells the bounds of the _pixel corners_.

#### Collection 2 level 2 Landsat 8

```
> aws s3 cp s3://usgs-landsat/collection02/level-2/standard/oli-tirs/2020/072/086/LC08_L2SP_072086_20201203_20201218_02_T1/LC08_L2SP_072086_20201203_20201218_02_T1_MTL.json - --request-payer \
    | jq '.LANDSAT_METADATA_FILE.PROJECTION_ATTRIBUTES | {CORNER_UL_PROJECTION_X_PRODUCT, CORNER_UL_PROJECTION_Y_PRODUCT, CORNER_LR_PROJECTION_X_PRODUCT, CORNER_LR_PROJECTION_Y_PRODUCT}'
{
  "CORNER_UL_PROJECTION_X_PRODUCT": "367200.000",
  "CORNER_UL_PROJECTION_Y_PRODUCT": "-4030500.000",
  "CORNER_LR_PROJECTION_X_PRODUCT": "600900.000",
  "CORNER_LR_PROJECTION_Y_PRODUCT": "-4265100.000"
}
```

```
AWS_REQUEST_PAYER="requester" \
  rio cogeo info \
  s3://usgs-landsat/collection02/level-2/standard/oli-tirs/2020/072/086/LC08_L2SP_072086_20201203_20201218_02_T1/LC08_L2SP_072086_20201203_20201218_02_T1_SR_B2.TIF

Geo
    Crs:              EPSG:32660
    Origin:           (367185.0, -4030485.0)
    Resolution:       (30.0, -30.0)
    BoundingBox:      (367185.0, -4265115.0, 600915.0, -4030485.0)
```

#### Collection 2 level 1 Landsat 5

```
> aws s3 cp s3://usgs-landsat/collection02/level-1/standard/tm/1997/212/200/LT05_L1GS_212200_19970429_20200910_02_T2/LT05_L1GS_212200_19970429_20200910_02_T2_MTL.json - --request-payer \
    | jq '.LANDSAT_METADATA_FILE.PROJECTION_ATTRIBUTES | {CORNER_UL_PROJECTION_X_PRODUCT, CORNER_UL_PROJECTION_Y_PRODUCT, CORNER_LR_PROJECTION_X_PRODUCT, CORNER_LR_PROJECTION_Y_PRODUCT}'

{
  "CORNER_UL_PROJECTION_X_PRODUCT": "679800.000",
  "CORNER_UL_PROJECTION_Y_PRODUCT": "2666100.000",
  "CORNER_LR_PROJECTION_X_PRODUCT": "920100.000",
  "CORNER_LR_PROJECTION_Y_PRODUCT": "2451600.000"
}
```

```
> AWS_REQUEST_PAYER="requester" \
  rio cogeo info \
  s3://usgs-landsat/collection02/level-1/standard/tm/1997/212/200/LT05_L1GS_212200_19970429_20200910_02_T2/LT05_L1GS_212200_19970429_20200910_02_T2_B5.TIF

Geo
    Crs:              EPSG:32652
    Origin:           (679785.0, 2666115.0)
    Resolution:       (30.0, -30.0)
    BoundingBox:      (679785.0, 2451585.0, 920115.0, 2666115.0)
```

#### Collection 1:

Consider `LC08_L1TP_196028_20200903_20200903_01_RT` in the old `landsat-pds` bucket:

```
> aws s3 cp s3://landsat-pds/c1/L8/196/028/LC08_L1TP_196028_20200903_20200903_01_RT/LC08_L1TP_196028_20200903_20200903_01_RT_MTL.json - \
	| jq '.L1_METADATA_FILE.PRODUCT_METADATA | {CORNER_UL_PROJECTION_X_PRODUCT, CORNER_UL_PROJECTION_Y_PRODUCT, CORNER_LR_PROJECTION_X_PRODUCT, CORNER_LR_PROJECTION_Y_PRODUCT}'

{
  "CORNER_UL_PROJECTION_X_PRODUCT": 609000,
  "CORNER_UL_PROJECTION_Y_PRODUCT": 5217300,
  "CORNER_LR_PROJECTION_X_PRODUCT": 838500,
  "CORNER_LR_PROJECTION_Y_PRODUCT": 4984200
}
```

```
> rio cogeo info s3://landsat-pds/c1/L8/196/028/LC08_L1TP_196028_20200903_20200903_01_RT/LC08_L1TP_196028_20200903_20200903_01_RT_B2.TIF

Geo
    Crs:              EPSG:32631
    Origin:           (608985.0, 5217315.0)
    Resolution:       (30.0, -30.0)
    BoundingBox:      (608985.0, 4984185.0, 838515.0, 5217315.0)
```

You can see that even for the pan band, the bounding box lines up exactly with a half-pixel offset (here 7.5m).
```
> rio cogeo info s3://landsat-pds/c1/L8/196/028/LC08_L1TP_196028_20200903_20200903_01_RT/LC08_L1TP_196028_20200903_20200903_01_RT_B8.TIF

Geo
    Crs:              EPSG:32631
    Origin:           (608992.5, 5217307.5)
    Resolution:       (15.0, -15.0)
    BoundingBox:      (608992.5, 4984192.5, 838507.5, 5217307.5)
```

stactools_landsat/stactools/landsat/utils.py
kylebarron: Yeah I clearly missed that the first time 😄

matthewhanson:
Was going to respond, but you figured it out @kylebarron   The coordinates given in the MTL are the same as provided in the file, but they are the bounds of the entire image, nodata and everything.   It's the bounding box because it's a north-up image.
But the actual data footprint coordinates aren't available in the MTL.

stactools_landsat/stactools/landsat/utils.py
kylebarron: It looked like the pan band is similarly offset by half a pixel (7.5m) compared to the MTL bounds in the local CRS.

alexgleith: Ah, and the MTL I downloaded from here doesn't have `UTM_ZONE`.

https://landsatlook.usgs.gov/data/collection02/level-2/standard/oli-tirs/2020/081/119/LC08_L2SR_081119_20200101_20200823_02_T2/LC08_L2SR_081119_20200101_20200823_02_T2_MTL.json

I'll check another.

alexgleith: S3 command to get it:
```
aws s3 cp --request-payer requester \
	s3://usgs-landsat/collection02/level-2/standard/oli-tirs/2020/081/119/LC08_L2SR_081119_20200101_20200823_02_T2/LC08_L2SR_081119_20200101_20200823_02_T2_MTL.json .
```

This file is `L2SR` whereas yours was `L2SP`. Surface reflectance vs. ???

I don't know why they're different...

kylebarron: That MTL doesn't have a UTM zone because `MAP_PROJECTION` is `PS`, i.e. Polar Stereographic. I believe every MTL where the projection is `UTM` includes a `UTM_ZONE` key. I haven't worked with the PS projection before, so I'm not certain how to handle such MTL files. But I assume that there are only two relevant PS projections, one for the north pole and one for the south pole, and should be obvious which one from the latitude values.

```bash
> cat LC08_L2SR_081119_20200101_20200823_02_T2_MTL.json| jq '.LANDSAT_METADATA_FILE.PROJECTION_ATTRIBUTES'
```

```json
{
  "MAP_PROJECTION": "PS",
  "DATUM": "WGS84",
  "ELLIPSOID": "WGS84",
  "VERTICAL_LON_FROM_POLE": "0.00000",
  "TRUE_SCALE_LAT": "-71.00000",
  "FALSE_EASTING": "0",
  "FALSE_NORTHING": "0",
  "GRID_CELL_SIZE_REFLECTIVE": "30.00",
  "REFLECTIVE_LINES": "7671",
  "REFLECTIVE_SAMPLES": "7761",
  "ORIENTATION": "NORTH_UP",
  "CORNER_UL_LAT_PRODUCT": "-82.03706",
  "CORNER_UL_LON_PRODUCT": "98.46113",
  "CORNER_UR_LAT_PRODUCT": "-79.92551",
  "CORNER_UR_LON_PRODUCT": "96.67232",
  "CORNER_LL_LAT_PRODUCT": "-81.46766",
  "CORNER_LL_LON_PRODUCT": "112.64692",
  "CORNER_LR_LAT_PRODUCT": "-79.47133",
  "CORNER_LR_LON_PRODUCT": "108.16484",
  "CORNER_UL_PROJECTION_X_PRODUCT": "857100.000",
  "CORNER_UL_PROJECTION_Y_PRODUCT": "-127500.000",
  "CORNER_UR_PROJECTION_X_PRODUCT": "1089900.000",
  "CORNER_UR_PROJECTION_Y_PRODUCT": "-127500.000",
  "CORNER_LL_PROJECTION_X_PRODUCT": "857100.000",
  "CORNER_LL_PROJECTION_Y_PRODUCT": "-357600.000",
  "CORNER_LR_PROJECTION_X_PRODUCT": "1089900.000",
  "CORNER_LR_PROJECTION_Y_PRODUCT": "-357600.000"
}
```

alexgleith: And now I feel a bit silly, @kylebarron! I haven't encountered Landsat scenes in polar stereographic before... I picked a bad example! hehe

Although it does demonstrate the extra case that needs to be handled in translating MTLs.

kylebarron: I haven't worked with PS scenes either. I'm guessing you'd just need an if/else to switch between georeferencing for UTM and PS. But this should allow creating any image transform without reading the image I think (though I've already been wrong once 😄 )

alexgleith: Yeah, we can add handling of the (presumably) two polar projections and many UTM zones separately... I'll find some example STAC files and add more tests this week.

stactools_landsat/stactools/landsat/utils.py
alexgleith: MTL aside, I don't think there's CRS info in the STAC file either, which is a bit of an issue really.

Inferring CRS from the bounding box of the original STAC is fraught with danger, I think...

stactools_landsat/stactools/landsat/commands.py
alexgleith: I changed to using pystac.Item.from_file, which works fine. Using pystac.from_file does _not_ work, but I don't want to dive into why currently!

tests/data-files/landsat/readme.md
alexgleith: It's not needed, but is pretty convenient!

You sure you want it gone?

alexgleith:
I think I've resolved all the issues here. Assuming CI is happy, I'm happy!

stactools_landsat/stactools/landsat/utils.py
lossyrob: Followup to this conversation captured in https://github.com/stac-utils/stactools/issues/48 - feel free to correct or add to that issue!

lossyrob (approved):
LGTM, thanks!

PR #52, opened 2021-02-01
Add Sentinel 2 item creation command

alkamin:
## Overview
This PR adds a `sentinel` module to the CLI with a `create-item` command that will create a STAC item from a Sentinel2 L2A SAFE granule. This covers granules provided by ESA and any produced by `sen2cor`. It can optionally create COGs for any non-COG geo-referenced assets (in this case JP2 and TIFF assets) and includes them in the newly created item along-side the non-COG imagery.

Each item implements the `sat` and `eo` extensions and pulls in a substantial amount of metadata from the various XML files provided with the L2A granules.

The item provides the following metadata from the common metadata spec and the `eo` and `sat` extensions:

```json
        "constellation": "Sentinel 2",
        "platform": "Sentinel-2B",
        "instruments": [
            "msi"
        ],
        "eo:cloud_cover": 0.447807,
        "sat:relative_orbit": 81,
        "sat:orbit_state": "DESCENDING"
```

And then also provides the following metadata specific to S2:

```json
        "s2:generationTime": "2020-10-07T16:08:57.135Z",
        "s2:datatakeIdentifier": "GS2A_20190212T192651_019029_N02.12",
        "s2:datastripIdentifier": "S2A_OPER_MSI_L2A_DS_ESRI_20201007T160858_S20190212T192646_N02.12",
        "s2:granuleIdentifier": "S2A_OPER_MSI_L2A_TL_ESRI_20201007T160858_A019029_T07HFE_N02.12",
        "s2:datatakeType": "INS-NOBS",
        "s2:mgrsTile": "07HFE",
        "s2:processingBaseline": "02.12",
        "s2:productURI": "S2A_MSIL2A_20190212T192651_N0212_R013_T07HFE_20201007T160857.SAFE",
        "s2:productType": "S2MSI2A",
        "s2:cloudCoverageAssessment": 51.580326,
        "s2:meanSolarZenith": 32.707073851362,
        "s2:meanSolarAzimuth": 62.3286549448294,
        "s2:reflectanceConversionFactor": 1.02763689829235,
        "s2:meanIncidenceZenithAngleB01": 10.8927842996721,
        "s2:meanIncidenceAzimuthAngleB01": 290.48823382165,
        "s2:meanIncidenceZenithAngleB09": 10.9219716766936,
        "s2:meanIncidenceAzimuthAngleB09": 290.912974638118,
        "s2:meanIncidenceZenithAngleB10": 10.7681196122624,
        "s2:meanIncidenceAzimuthAngleB10": 288.149931950428,
        "s2:meanIncidenceZenithAngleB02": 10.7227053564162,
        "s2:meanIncidenceAzimuthAngleB02": 286.820822640182,
        "s2:meanIncidenceZenithAngleB03": 10.7484644698818,
        "s2:meanIncidenceAzimuthAngleB03": 287.66193062313,
        "s2:meanIncidenceZenithAngleB04": 10.778802813877,
        "s2:meanIncidenceAzimuthAngleB04": 288.433166003466,
        "s2:meanIncidenceZenithAngleB05": 10.7978110871057,
        "s2:meanIncidenceAzimuthAngleB05": 288.845299081921,
        "s2:meanIncidenceZenithAngleB06": 10.819076646934,
        "s2:meanIncidenceAzimuthAngleB06": 289.271947026448,
        "s2:meanIncidenceZenithAngleB07": 10.8418686954573,
        "s2:meanIncidenceAzimuthAngleB07": 289.683442309859,
        "s2:meanIncidenceZenithAngleB08": 10.734636981112,
        "s2:meanIncidenceAzimuthAngleB08": 287.241471003298,
        "s2:meanIncidenceZenithAngleB8A": 10.8667503672554,
        "s2:meanIncidenceAzimuthAngleB8A": 290.104191276527,
        "s2:meanIncidenceZenithAngleB11": 10.8148807894454,
        "s2:meanIncidenceAzimuthAngleB11": 289.168163140243,
        "s2:meanIncidenceZenithAngleB12": 10.8717150734,
        "s2:meanIncidenceAzimuthAngleB12": 290.159788183946,
        "s2:solarIrradianceB01": 1884.69,
        "s2:solarIrradianceB02": 1959.72,
        "s2:solarIrradianceB03": 1823.24,
        "s2:solarIrradianceB04": 1512.06,
        "s2:solarIrradianceB05": 1424.64,
        "s2:solarIrradianceB06": 1287.61,
        "s2:solarIrradianceB07": 1162.08,
        "s2:solarIrradianceB08": 1041.63,
        "s2:solarIrradianceB8A": 955.32,
        "s2:solarIrradianceB09": 812.92,
        "s2:solarIrradianceB10": 367.15,
        "s2:solarIrradianceB11": 245.59,
        "s2:solarIrradianceB12": 85.25,
```

I've added three down-sampled Sentinel2 L2A granules from different sources to `tests/data-files` which can be used to test drive the command without incurring a major time-cost.

## Notes

`Product_URI` seemed like the best field to use as an ID as it should be consistent across providers

## Testing
1) `docker/console`
2) to test without cogification: `scripts/stac sentinel create-item ./tests/data-files/sentinel2/S2A_MSIL2A_20190212T192651_N0212_R013_T07HFE_20201007T160857.SAFE/ .
3) to test with cogification you can add the `-c` or`--cogify` flag


stactools_sentinel/README.md
```diff
@@ -0,0 +1,5 @@
+# stactools sentinel
+
+A subpackage of stactools for working with Sentinel data.
+
+__WORK IN PROGRESS__. There is no functionality in this libary yet.
```
alkamin, line 5: Not really true, will update.

stactools_sentinel/stactools/sentinel/stac.py
```diff
@@ -0,0 +1,349 @@
...
+    item.add_asset(*datastrip_mtd_asset)
+    item.add_asset(*granule_mtd_asset)
+    item.add_asset(*thumbnail_asset)
+
+    for asset in image_assets:
+        item.add_asset(*asset)
+
+    item.ext.eo.cloud_cover = cloudy_percentage
+
+    # TODO: when pystac SAT extension support is fixed, do this addition through the extension
```
alkamin, line 162: https://github.com/stac-utils/pystac/issues/257

matthewhanson:
Hi @alkamin , thanks for this, it's been on my list of things to do but never gotten to it.

Did you use the stac-sentinel as a reference at all for this?
https://github.com/stac-utils/stac-sentinel/blob/master/stac_sentinel/sentinel.py#L20

Which is what was used to create all of the Sentinel-2 catalogs on Earth-search: https://earth-search.aws.element84.com/v0

I think it's helpful to have a few key sentinel specific tags (grid_square, latitude_band, sequence_id, product_id), but I'm not so sure all of the Sentinel-2 fields need to be carried over, since the original S2 metadata can be included as an asset.

Going to take a pass through the commits and leave comments there as well.

One thing we haven't discussed @lossyrob is that if stactools is providing mappings for various datasets, reference collection level metadata should also be included here.
Additionally if there are vendor specific fields, the JSON schema for that should also be included and referenced from stac_extensions.

matthewhanson:
Do we need to have an example SAFE archive in here?  It adds a lot of files, and can really bloat the repo. 

Maybe a process/script  to fetch test data when testing instead of including it by default?

stactools_sentinel/stactools/sentinel/constants.py
```diff
@@ -0,0 +1,92 @@
...
+                full_width_half_max=0.045),
+    'B04':
+    Band.create(name='B04',
+                common_name='red',
+                description='Band 4 - Red',
+                center_wavelength=0.665,
+                full_width_half_max=0.038),
+    'B05':
+    Band.create(name='B05',
+                common_name='rededge1',
```
matthewhanson, line 46: These common_names (rededgeX, watervapor) are not part of the STAC list, and don't follow the convention of naming.
It would be nice to revisit that table and add some additional bands, but the red edge bands are more difficult because of how narrow they are and unique they tend to be for each given instrument.
In the meantime, you can leave off common_name for any band that doesn't have one:
https://github.com/radiantearth/stac-spec/tree/master/extensions/eo#common-band-names

stactools_sentinel/stactools/sentinel/constants.py
```diff
@@ -0,0 +1,92 @@
...
+                full_width_half_max=0.033),
+    'B09':
+    Band.create(name='B09',
+                common_name='watervapor',
+                description='Band 9 - Water vapor',
+                center_wavelength=0.945,
+                full_width_half_max=0.026),
+    'B11':
+    Band.create(name='B11',
+                common_name='swir1',
```
matthewhanson, line 82: `swir1` is not a valid common_name, it should be `swir16` (representing the atmospheric window at 1.6 microns)

stactools_sentinel/stactools/sentinel/constants.py
```diff
@@ -0,0 +1,92 @@
...
+                description='Band 11 - SWIR 1',
+                center_wavelength=1.610,
+                full_width_half_max=0.143),
+    'B12':
+    Band.create(name='B12',
+                common_name='swir2',
```
matthewhanson, line 88: `swir1` is not a valid common_name, it should be `swir22` (representing the atmospheric window at 2.2 microns)

alkamin:
@matthewhanson at a minimum I'll remove the unused files from the SAFE archives

alkamin:
@matthewhanson Regarding the metadata stuff I agree and don't quite know what the right balance is to strike. Working with @lossyrob on this a bit and the goal was to figure out what metadata would actually be useful for analysis purposes. Based on what metadata GEE surfaces and how tedious parsing the XML data was I felt it would be better to err on the side of more metadata to make it easier for item consumers. Sometimes less is more though.

Following the same theme of reducing processing for the consumer, the other decision was to include assets for all available resolutions for each band. It does result in _a lot_ of assets though so it's another thing to consider.

I don't think I pulled out latitude_band, and sequence_id though so I can do that.

alkamin:
Update on this:

Latest push gets lint removed and also adds a `scripts/lint` script that can be run without also having to run tests which is helpful. I've also added an additional flake8 rule to be ignored to the flake8 settings that was causing disagreement between yapf and flake8 regarding ending a line with a binary operator.

Regarding the metadata discussion, I've put together a proposal for an extended metadata sidecar file. The proposal is [here](https://gist.github.com/alkamin/3a261b731c98e45aeee2efe42feb7dd2). I've surfaced it in the STAC gitter channel as well. Once there is some feedback on the proposal I'll take a pass at implementing it within stactools so it can be evaluated in a functional state.

stactools_sentinel/stactools/sentinel/extended_item_metadata.py
```diff
@@ -0,0 +1,64 @@
+import pystac
+from pystac import (STACObject, Link)
+
+
+class ExtendedItemMetadata(STACObject):
```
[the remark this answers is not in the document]
lossyrob: Actually not seeing this used below, but there's an `item.clone()`...is this class now unused?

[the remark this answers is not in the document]
alkamin: Yeah, this class was not supposed to survive to the PR. 🗑️

TomAugspurger:
Lazy question (I haven't looked at the code yet): Does this capture the data needed for the project STAC extension (https://github.com/radiantearth/stac-spec/tree/master/extensions/projection)? This would be helpful for projects like stac-vrt: https://stac-vrt.readthedocs.io/en/latest/#building-stac-vrt-compatible-stac-items

stactools_sentinel2/stactools/sentinel2/utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+def orjson_save_json(uri, json_dict):
+    dirname = os.path.dirname(uri)
+    if dirname != '' and not os.path.isdir(dirname):
+        os.makedirs(dirname)
+    with open(uri, 'wb') as f:
+        f.write(orjson.dumps(json_dict, option=orjson.OPT_INDENT_2))
+
+
+def orjson_save_item(item, include_self_link=True, dest_href=None):
```
lossyrob, line 75: Does this work with remote paths?

Using the pystac save method takes advantage of fsspec (set via [here](https://github.com/stac-utils/stactools/blob/master/stactools_core/stactools/core/io.py#L5)) for STAC_IO, which allows users to plug in different file systems and write directly to remote storage. While the speedup is nice, if this only writes to local we'd want to make the usage of orjson is wrapped in an option at least.

lossyrob: Which reminds me that [this line](https://github.com/stac-utils/stactools/blob/master/stactools_naip/stactools/naip/__init__.py#L7) is missing from the subpackage; can you put that in and then make a note in the README section about adding new datasets that this is one of the steps? This is required to enable fsspec usage in the subpackage.

lossyrob:
@alkamin I put in the last docstring, this is good to go, I'm going to merge. Thanks!

PR #53, opened 2021-02-05
Fixed rasterio building from source

TomAugspurger:
Disables pip's build isolation to ensure that rasterio is built against
the same NumPy we're using in this environment.


TomAugspurger:
cc @lossyrob.

lossyrob (approved):

PR #54, opened 2021-02-06
Add method for creating MODIS.

pflickin:


lossyrob:
This will have to be migrated to `stactools-packages`; closing for now pending that work.

gadomski:
Done here: https://github.com/stactools-packages/modis. Note that CI will be failing so it will require touch-up.

PR #61, opened 2021-02-15
add catalog type option for 'planet convert'

### LINKED ISSUE #29: Option to specify catalog type in more places
This is probably a useful thing for other 'convert' commands, but this is the one I'm using the most. 

Most of my workflows involve eventually sending the catalog out, either uploading to google cloud or putting into a zip to send someone. It'd be nice if I could just continually work with self-contained catalogs, instead of having to remember to do a copy with `-t`. This hits me most with `planet convert`, but also probably in merging (or my guess is it might).

volaya:
fixes #29

### DIFF
--- a/stactools_planet/stactools/planet/commands.py
+++ b/stactools_planet/stactools/planet/commands.py
@@ -16,7 +16,8 @@ def convert_order(manifest,
                   copy=False,
                   description=None,
                   title=None,
-                  skip_validation=False):
+                  skip_validation=False,
+                  catalog_type=None):
     manifest = OrderManifest.from_file(manifest)
     collection = manifest.to_stac(collection_id=collection_id,
                                   description=description,
@@ -28,7 +29,8 @@ def convert_order(manifest,
 
     logger.info('Saving STAC collection at {}...'.format(
         collection.get_self_href()))
-    collection.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
+    catalog_type = catalog_type or pystac.CatalogType.SELF_CONTAINED
+    collection.save(catalog_type=catalog_type)
 
     # Move assets after save to avoid moving without saving,
     # and so the directory structure is generated.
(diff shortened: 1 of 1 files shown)


lossyrob:
Looks good to me, just need to format the code to pass CI

stactools_planet/stactools/planet/commands.py
```diff
@@ -16,7 +16,8 @@ def convert_order(manifest,
                   copy=False,
                   description=None,
                   title=None,
-                  skip_validation=False):
+                  skip_validation=False,
+                  catalog_type=None):
```
kylebarron, line 20: You could just default to `pystac.CatalogType.SELF_CONTAINED` and remove the null check below?

stactools_planet/stactools/planet/commands.py
```diff
@@ -60,8 +62,16 @@ def planet():
...
                   help='Optional title for the STAC collection.')
     @click.option('--skip-validation', help='Skip validation on the STAC')
+    @click.option('-c',
+                  '--catalog-type',
+                  type=click.Choice([
+                                    pystac.CatalogType.ABSOLUTE_PUBLISHED,
+                                    pystac.CatalogType.RELATIVE_PUBLISHED,
+                                    pystac.CatalogType.SELF_CONTAINED
+                                    ],
+                                    case_sensitive=False))
```
kylebarron, line 72: I was going to suggest just [iterating over the enum](https://stackoverflow.com/a/29503454), like `[e.value for e in pystac.CatalogType]`, but then I learned that `pystac.CatalogType` doesn't subclass from `Enum`. @lossyrob is there a reason you don't want `CatalogType` to subclass from `Enum`, or would you accept a PR? I'd have to double check that it would be backwards compatible but it should be.

kylebarron: I see this is https://github.com/stac-utils/pystac/issues/231

volaya: Notice that there are other places in the stactools cli where that list is used. I actually just took it from a different command that has the option to specify the type of catalog, so in case of having enums available, it should be changed not just here

lossyrob (approved):

PR #69, opened 2021-03-06
add file info to assets when creating catalog from Planet data

volaya:
This PR requires the changes in pystac implemented in https://github.com/stac-utils/pystac/pull/270

### DIFF
--- a/stactools_planet/stactools/planet/planet_item.py
+++ b/stactools_planet/stactools/planet/planet_item.py
@@ -2,12 +2,15 @@ import os
 from copy import deepcopy
 import json
 import logging
+import hashlib
 
+from multihash import Multihash
 import fsspec
 import pystac
 from pystac.link import Link
 from pystac.utils import (str_to_datetime, make_absolute_href)
 from pystac.extensions.eo import Band
+from pystac.extensions.file import FileDataType
 from shapely.geometry import shape
 
 from stactools.planet import PLANET_PROVIDER
@@ -57,6 +60,19 @@ class PlanetItem:
         self.item_assets = item_assets
         self.base_dir = make_absolute_href(base_dir)
         self.metadata_href = metadata_href
+
+    def _file_checksum(self, path):
+        BUF_SIZE = 65536
+        sha1 = hashlib.sha1()
+
+        with open(path, 'rb') as f:
+            while True:
+                data = f.read(BUF_SIZE)
+                if not data:
+                    break
+                sha1.update(data)
+                mh = Multihash.from_hash(sha1)
+                return mh.encode('hex').decode('ascii')
 
     def to_stac(self):
         props = deepcopy(self.item_metadata['properties'])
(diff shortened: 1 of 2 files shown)


stactools_planet/stactools/planet/planet_item.py
```diff
@@ -125,44 +141,46 @@ def to_stac(self):
...
+                        geotransform = dataset.transform
+                        if width > height:
+                            width, height = 256, int(height / width * 256)
+                        else:
+                            width, height = int(width / height * 256), 256
+
+                        profile = dataset.profile
+                        profile.update(driver='PNG')
+                        profile.update(width=width)
+                        profile.update(height=height)
```
kylebarron, line 165: FYI you can combine these into a single command if you want:
```py
>>> a = {}
>>> a.update(a=1, b=2)
>>> a
{'a': 1, 'b': 2}
```

stactools_planet/stactools/planet/planet_item.py
```diff
@@ -196,6 +210,17 @@ def to_stac(self):
                         ]
                     item.ext.eo.set_bands(bands, asset)
 
+            if add_file_info:
+                item.ext.enable('file')
+                checksum = self._file_checksum(href)
+                with rasterio.open(href) as dataset:
+                    # we assume all bands in the image have the same data type
+                    data_type = FileDataType(dataset.dtypes[0])
+                size = os.path.getsize(href)
```
kylebarron, line 219: `getsize` will only work, but using the name `href` to me implies it should work with remote paths too. Is that intended?

volaya: I am just the variable that was there before these changes, but you are right, that won't work with a remote file path. We can put that in a try block ad only add those parameters when it's possible.

In any case, I am not sure if href in this case can be a remote path. Isn't the idea of this extension to create a STAC catalog from a planet bundle that you have downloaded from Planet (and is therefore local)?

stactools_planet/stactools/planet/planet_item.py
volaya: good catch. Will do it

stactools_planet/stactools/planet/planet_item.py
kylebarron: > Isn't the idea of this extension to create a STAC catalog from a planet bundle that you have downloaded from Planet (and is therefore local)? 

I haven't used this module, but I figured it might have also been intended to work with a Planet API URL. From a cursory look, other steps like the `rasterio.open` would work with both local files and remote URLs.

gadomski:
@volaya just checking to see if this PR is still something that we'd like to add to Planet. If so, we'll want to move these changes over to the new https://github.com/stactools-packages/planet as a part of the v0.2 reorganization. Let me know what you'd like to do and we can work together to make that happen. Thanks!

volaya:
@gadomski Sorry for the late reply. Here I have merged the changes in this branch over the current planet extension repo that you linked above. 

https://github.com/volaya/planet/commit/fb944f4cfe433b272f457a6fa5ec962ed326207f

Probably more work is done to adapt to the current pystac api, since extensions are not handled the same way, as far as I have seen...

Ignore the commit I just made in this PR...I thought I was pushing to another repo and I pushed to this one. Probably this PR can be closed, and we can create a new one with that commit that contains the same changes but applied to the other repo

gadomski:
> Probably this PR can be closed, and we can create a new one with that commit that contains the same changes but applied to the other repo

Sounds good, can you open that PR when it's feeling ready?

> Probably more work is done to adapt to the current pystac api, since extensions are not handled the same way, as far as I have seen...

I have a work in progress branch for PySTAC v1.0 support in the planet package here: https://github.com/stactools-packages/planet/tree/feature/pystac-v1.0. https://github.com/stac-utils/stactools/pull/113 is the PySTAC v1.0 support in stactools which is the base for the planet branch.

Issue #44, opened 2021-01-21
Convert pointcloud .las to STAC


cholmes:
It'd be great to have a stactool that can convert from a pointcloud in the las format to a STAC item. And perhaps a way to handle a directory of point clouds and convert that to a catalog. May also be interesting to have an explicit 3dep one, if it has particular metadata.

gadomski:
https://github.com/stactools-packages/pointcloud has been created, but is not yet a "core package." We don't have a procedure to onboard new core packages, #140 outlines some of the considerations. Once those are resolved, we should try to onboard `pointcloud`.

gadomski:
"Core packages" never really took off, so closing as OBE.

PR #71, opened 2021-03-12
Create STAC Item for .las (and .laz) pointclouds

gadomski:
**Related issue(s):** stac-utils/stactools#44

**Description:** API function and CLI tool to create STAC Items from pointclouds. We're using [PDAL](https://pdal.io), which will enable us to (eventually) support formats other than la{s|z}. However, right now the STAC item is built from fields from the las header, so other formats are explicitly denied for now.


hobu:
> Another option to get network-aware las reading is to dig through the other Python las implementations and see if one is good with networks (or write Yet Another Las Header Reader)

This code lives in Entwine. We could put it to the ``preview()`` method of PDAL I suppose. Let's not write another Python LAS library 😄

gadomski:
Turns out I'm dumb and the ﻿PDAL Python reader [does support network access](https://pypi.org/project/PDAL/#reading-using-numpy-arrays). So this is ready for review.

gadomski:
Closed in favor of https://github.com/stactools-packages/stactools-pointcloud as a part of #111.

PR #81, opened 2021-03-26
Add support for USGS 3DEP (formerly NED) items and collections

gadomski:
3DEP is the [3D Elevation Program from USGS](https://www.usgs.gov/core-science-systems/ngp/3dep). This commit provides support for creating items and collections for 3DEP data hosted at ftp://rockyftp.cr.usgs.gov/vdelivery/Datasets/Staged/Elevation.

Also includes two linty commits relating to a `cgls_lc100` dangling symlink and old version string.


kylebarron:
> This commit provides support for creating items and collections for 3DEP data hosted at ftp://rockyftp.cr.usgs.gov/vdelivery/Datasets/Staged/Elevation.

Does it also work with the same data (I think) in the `prd-tnm` S3 bucket? Everything under `StagedProducts/Elevation/`

https://s3.console.aws.amazon.com/s3/buckets/prd-tnm?region=us-west-2&prefix=StagedProducts/Elevation/&showversions=false

gadomski:
> Does it also work with the same data (I think) in the prd-tnm S3 bucket? 

Yep, 4ff1abbc35ab096e73a35665769d7d3a057f1781 switches to using the `AWS_BASE` (https://prd-tnm.s3.amazonaws.com/StagedProducts/Elevation) as the default. It's built so that AWS, the USGS FTP, and arbitrary other bases are hopefully interchangeable.

rsignell-usgs:
@gadomski and @kylebarron , just discovered this, thanks to @TomAugspurger !  
The USGS is eager to create STAC items that will work with stackstac and stac-vrt for this dataset, so just let me know what the next step should be on our end!

gadomski:
@rsignell-usgs right on!

> create STAC items that will work with stackstac and stac-vrt for this dataset

**stac-vrt** has some [reccommended attributes](https://stac-vrt.readthedocs.io/en/latest/#building-stac-vrt-compatible-stac-items) which [we are including](https://github.com/stac-utils/stactools/pull/81/files#diff-1512e546de988f1bc012b31893b48df11da5a2e8ac00f5f89a44c6cb493c0772R145-R177). I'm not very familiar with **stackstac** so am less sure what STAC attributes are required/recommended there, and wasn't able to find a quick reference w/ that info. @gjoseph92 would you be able to provide guidance on any optional attributes to include in our STAC objects to best support **stackstac**?

stactools_threedep/stactools/threedep/commands.py
```diff
@@ -0,0 +1,126 @@
...
+                title = "1/3 arc-second"
+                description = "USGS 3DEP 1/3 arc-second DEMs"
+            else:
+                raise NotImplementedError
+            catalog = Catalog(id=product, description=description, title=title)
+            collection.add_child(catalog)
+            catalog.add_items(items[product])
+        collection.update_extent_from_items()
+        collection.normalize_hrefs(destination)
+        collection.save(catalog_type=CatalogType.SELF_CONTAINED)
```
matthewhanson, line 77: With the recent change in STAC spec, Collection does not inherit from Catalog, which means that a Collection cannot act as a root catalog.
It looks like here a top-level Collection is created and saved without a root. 

I'm not sure if this is common across other stactools modules, but I think the workflow should for creating any new collection should first assume an existing root catalog, and new collections are added to that.

stactools_threedep/stactools/threedep/metadata.py
```diff
@@ -0,0 +1,218 @@
...
+                     media_type=MediaType.XML,
+                     roles=["metadata"])
+
+    def thumbnail_asset(self, base: str = DEFAULT_BASE) -> Asset:
+        """Returns the numbnail asset."""
+        return Asset(href=self._asset_href_with_extension(base, "jpg"),
+                     media_type=MediaType.JPEG,
+                     roles=["thumbnail"])
+
+    def derived_from_link(self, base: str = DEFAULT_BASE) -> Link:
```
matthewhanson, line 139: The intention of the `derived_from` link was that it allows for tracing provenance through STAC Items. In other words, the target of `derived_from` should be another STAC Item.
To point back to the original metadata that was used to create the STAC Item, use `via`:
https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#using-relation-types

stactools_threedep/stactools/threedep/metadata.py
```diff
@@ -0,0 +1,218 @@
...
+        }
+        geometry = mapping(box(*bbox))
+        return {
+            "epsg": THREEDEP_EPSG,
+            "shape": shape,
+            "transform": transform,
+            "bbox": bbox,
+            "centroid": centroid,
+            "geometry": geometry,
+        }
```
matthewhanson, line 177: The `projection` extension does have a few fields that I think should always be included in STAC metadata as it allows apps like stac-vrt, stackstac and OpenDataCube to do their magic without first hitting data files. 
These fields include `epsg`, `shape`, and `transform`.

However, I don't see much use for bbox in native coordinates (this can be calculated from transform) at all.  `centroid` and `geometry` may be useful for some small number of use cases, but I think they are generally unnecessary.

thoughts @lossyrob ?

gadomski: If I recall correctly I included `bbox`, `centroid`, and `geometry` because [`Extension.apply`](https://github.com/stac-utils/stactools/pull/81/files#diff-880b7ec34b4338db8c7e89e38252fcbe043576a8382c13db6a619ae47969792cR47) was complaining without them, and they weren't too expensive to provide. Need to check that, and if that's true maybe it make sense to relax `apply`.

stactools_threedep/stactools/threedep/stac.py
```diff
@@ -0,0 +1,48 @@
...
+                geometry=geometry,
+                bbox=bbox,
+                datetime=metadata.datetime,
+                properties={})
+    start_datetime = metadata.start_datetime
+    end_datetime = metadata.end_datetime
+    if start_datetime and end_datetime:
+        item.common_metadata.start_datetime = start_datetime
+        item.common_metadata.end_datetime = end_datetime
+    item.common_metadata.gsd = metadata.gsd
```
matthewhanson, line 41: I'm sure you considered the other common metadata parameters, but I was wondering if there was any info about the specific platform or instrument used to collect any given piece of data.
Seems like `mission` could be 3dep, but it's a bit redundant since the collection already encapsulated the whole mission.

https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#instrument

gadomski: I'll give it another passthrough, but in general it doesn't seem to make sense to provide more specific platform information since the data is relatively mixed-provenance. If we end up supporting focus-region data (e.g. 1m DEMs and such) then platform and friends should make more sense.

gjoseph92:
@gadomski FYI there's some info on this buried in the documentation for the `resolution` argument https://stackstac.readthedocs.io/en/latest/api/main.html, but basically stackstac just wants the same things as stac-vrt (the important fields from the `proj` extension).

stactools_threedep/stactools/threedep/commands.py
gadomski: Screen Shot 2021-04-05 at 2 30 51 PM

FYI I think my confusion arose from [here](https://github.com/radiantearth/stac-spec/blob/master/catalog-spec/catalog-spec.md#link-object), where it says that `root` can point to a `Catalog` or a `Collection`.

I will still fix to implement `Catalog -> 2xCollection (one for 1 arc-second and one for 1/3 arc second -> Items`, with maybe some sort of subsetting catalogs to group the items better as well.

matthewhanson: Ah, good catch.  That should be updated as a consequence of not having Collections inherit from Catalogs.
However in looking through the latest stac-spec version I'm having trouble finding the changes that were made to that wording.  Will open up issue

matthewhanson: I'm still not convinced that there should be two collections here, vs one collection where you would be able to filter on `gsd`.  We should get some additional feedback on that.

gadomski: Roger. I guess it feels weird to me then to have a `1 -> 1 -> many` tree -- feels like there's an unnecessary node.

matthewhanson: You mean if it was Catalog->Collection->Items?
You'd normally have other non-3DEP collections in your Catalog, it also mirrors a STAC API where there's a root catalog with Collections under it.

stactools_threedep/stactools/threedep/metadata.py
matthewhanson: There should only be a problem if they are required fields. If that's the case, should open up issue in PySTAC

stactools_threedep/stactools/threedep/commands.py
gadomski: Gotcha. Yeah, I guess I'm thinking about if we expand this function to create a `Catalog` that encompasses more than just the 1 and 1/3 arc second data, e.g. the 1m which has less complete coverage. In that case, my thinking is we'd want a separate `Collection` for the 1m at least, so that its bounds are tighter than the 1 arc-second. Or am I overusing `Collection`s?

matthewhanson (approved):

PR #91, opened 2021-04-28
Fixed ST_B10 band

carioca-au:
In my last fix, I commit the mistake to add B10 as SR. Fixing that and also adding an exception treatment.


gadomski:
Closing in favor of #100.

PR #96, opened 2021-04-29
Added ST_B6 as fallback to B10

carioca-au:
Added B6 as a fallback to B10. It was necessary due B10 was not found within Landsat 5 ST stac 0.7

### DIFF
--- a/stactools_landsat/stactools/landsat/utils.py
+++ b/stactools_landsat/stactools/landsat/utils.py
@@ -108,8 +108,11 @@ def transform_stac_to_stac(item: Item,
             # ST_B10.TIF is a fallback for SR_B2
             elif item.assets.get("ST_B10.TIF"):
                 asset = item.assets["ST_B10.TIF"]
+            # Landsat5 does not have B10
+            elif item.assets.get("ST_B6.TIF"):
+                asset = item.assets["ST_B6.TIF"]
             else:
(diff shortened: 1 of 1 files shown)


gadomski:
After looking at this PR and #91, is there any reason not to enumerate all possible assets w/ proj information (with some pre-defined order) to extract the information? E.g.

```python
proj_dict = None
for asset_tag in PROJ_ASSET_TAGS:
    asset = item.assets[asset_tag]
    if asset:
         proj_dict = make_proj_dict(asset)
         break
if proj_dict:
    item.ext.projection.apply(**proj_dict)
```

or whatever. Looking to avoid another PR to add another fallback if ST_B6 isn't present.

carioca-au:
> After looking at this PR and #91, is there any reason not to enumerate all possible assets w/ proj information (with some pre-defined order) to extract the information? E.g.
> 
> ```python
> proj_dict = None
> for asset_tag in PROJ_ASSET_TAGS:
>     asset = item.assets[asset_tag]
>     if asset:
>          proj_dict = make_proj_dict(asset)
>          break
> if proj_dict:
>     item.ext.projection.apply(**proj_dict)
> ```
> 
> or whatever. Looking to avoid another PR to add another fallback if ST_B6 isn't present.

I reckon you are right, I'll change the code and create a new PR... Thank you

PR #100, opened 2021-05-03
Use any asset to retrieve projection information for Landsat items

carioca-au:
- dropping the idea to test just blue band geotiff files, the information that rasterio needs can come from any geotiff file, so getting from the first listed.
 - Fixed test


stactools_landsat/stactools/landsat/utils.py
```diff
@@ -102,40 +103,33 @@ def transform_stac_to_stac(item: Item,
...
+                if "geotiff" in asset.media_type:
+                    if not retrieved_tile_info:
+                        with rasterio.open(asset.href) as opened_asset:
+                            shape = opened_asset.shape
+                            transform = opened_asset.transform
+                            crs = opened_asset.crs.to_epsg()
+                            retrieved_tile_info = True
+
                     item.ext.projection.set_transform(transform, asset=asset)
                     item.ext.projection.set_shape(shape, asset=asset)
```
gadomski, line 117: It looks like the projection extension is being set multiple times here (one for each asset that has "geotiff" in the media type). This should probably be moved below this loop and only set once (i.e. to https://github.com/stac-utils/stactools/pull/100/files/8d3c80a8bfb290317f96637a58e6ed2d7f4fa412#diff-408a908da00c6d195d2d9215a17f6d21d27b3e39e63c4a017b4e562727356363R122).

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -102,40 +103,33 @@ def transform_stac_to_stac(item: Item,
...
-            print("Failed to load blue band, so not handling proj fields")
+            # Now we have the info, we can make the fields
+            item.ext.enable("projection")
+            item.ext.projection.epsg = crs
 
-    # Remove .TIF from asset names
-    new_assets = {}
+        except RasterioIOError as io_error:
+            print("Failed loading geotiff, so not handling proj fields")
+            raise io_error
```
gadomski, line 126: This `try/except` block is too big, this can just be around the `rasterio.open` call.

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -102,40 +103,33 @@ def transform_stac_to_stac(item: Item,
...
+
                     item.ext.projection.set_transform(transform, asset=asset)
                     item.ext.projection.set_shape(shape, asset=asset)
                     asset.media_type = MediaType.COG
 
-        except RasterioIOError:
-            print("Failed to load blue band, so not handling proj fields")
+            # Now we have the info, we can make the fields
+            item.ext.enable("projection")
+            item.ext.projection.epsg = crs
```
gadomski, line 122: This should raise an exception if `crs` (and `shape` and `transform`) are unset.

gadomski (changes requested):
In addition to the in-line comments/suggestions, can you add a regression test that shows an example of failure without this fix? Thanks!

stactools_landsat/stactools/landsat/utils.py
carioca-au: Hi Pete,

In this case 
```
shape = opened_asset.shape
transform = opened_asset.transform
crs = opened_asset.crs.to_epsg()
```
Will be set just the first time that the code finds a geotiff, it happens because of the flag ```retrieved_tile_info = True``` which won't allow to set that again. I added this code inside of the loop, because otherwise I'd have to loop twice.

stactools_landsat/stactools/landsat/utils.py
carioca-au: Pete,

It will raise an exception much before ```crs```. When the code tries to set ```shape``` and it's not defined it will raise the exception. 
What I can do and I'll do is add a better explanation(exception message) to this situation.

stactools_landsat/stactools/landsat/utils.py
carioca-au: Totally agree with you, I haven't changed due to the fact that, before my changes, the try/except was already around more than just rasterio. It'll be changed.

stactools_landsat/stactools/landsat/utils.py
carioca-au: Added new check, which raises an exception in case of missing any of those.

stactools_landsat/stactools/landsat/utils.py
carioca-au: Changed

carioca-au (commented):
All asked changes were done.
Thank you.

carioca-au:
> In addition to the in-line comments/suggestions, can you add a regression test that shows an example of failure without this fix? Thanks!

I've added a new test ```test_transform_static_stac_asset_not_available``` which will test the case if the asset isn't accessible.

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -101,49 +102,51 @@ def transform_stac_to_stac(item: Item,
...
+                            crs = opened_asset.crs.to_epsg()
+                            # Check to ensure that all information is present
+                            if not shape or not transform or not crs:
+                                raise STACError(f"Failed setting shape, transform and csr from {asset.href}")
+
+                    except RasterioIOError as io_error:
+                        raise STACError(f"Failed loading geotiff, so not handling proj fields, {io_error}")
+
+                item.ext.projection.set_transform(transform, asset=asset)
+                item.ext.projection.set_shape(shape, asset=asset)
```
gadomski, line 128: This still looks like we're setting the transform and the shape N times, where N is the number of assets. I think this should be indented one more level to be inside of the `if not shape:` block.

```suggestion
                    item.ext.projection.set_transform(transform, asset=asset)
                    item.ext.projection.set_shape(shape, asset=asset)
```

tests/landsat/test_commands.py
```diff
@@ -50,6 +54,20 @@ def test_transform_static_stac(self):
...
+            item = pystac.Item.from_file(stac_file)
+            for asset in item.get_assets():
+                # Replace a valid local file for an invalid url forcing an exception
+                asset.href = asset.href.replace(
+                    'tests/data-files/landsat/LC08_L2SR_081119_20200101_20200823_02_T2_SR_B2_small.TIF',
+                    "https://usgs.gov/just/an/invalid/url/LC08_L2SR_081119_20200101_20200823_02_T2_SR_B2_small.TIF"
+                )
+            with self.assertRaises(pystac.STACError):
+                transform_stac_to_stac(item)
+
```
gadomski, line 70: This is a good test, thanks for adding. I was also hoping to see one that shows that we do _not_ fail even though SR_B2 is missing, i.e. a test that would fail when run against the **master** branch, but would succeed against this branch.

gadomski (changes requested):
Looking good, thank you for the changes. Just two more to resolve.

tests/landsat/test_commands.py
carioca-au: Morning Pete,
I'll create a test removing B2 and B10 before the transformation, so it would break for master and pass on this branch.

stactools_landsat/stactools/landsat/utils.py
carioca-au: Morning Pete,
       in this case I'm not setting transform and shape to the item, I'm setting transform and shape to each asset and because of that it has to be inside of the loop.
       PYSATAC Library (projection.py):
                 asset.properties['proj:transform'] = transform => Line 425
                 asset.properties['proj:shape'] = shape => Line 379

tests/landsat/test_commands.py
carioca-au: @gadomski 
Test test_transform_static_stac_missing_asset_b2_b10 created!
I reckon now everything is right.

stactools_landsat/stactools/landsat/utils.py
```diff
@@ -119,10 +121,14 @@ def transform_stac_to_stac(item: Item,
...
+                                )
 
                     except RasterioIOError as io_error:
-                        raise STACError(f"Failed loading geotiff, so not handling proj fields, {io_error}")
+                        raise STACError(
+                            f"Failed loading geotiff, so not handling proj fields, {io_error}"
+                        )
 
                 item.ext.projection.set_transform(transform, asset=asset)
                 item.ext.projection.set_shape(shape, asset=asset)
```
gadomski:
```suggestion
                    item.ext.projection.set_transform(transform, asset=asset)
                    item.ext.projection.set_shape(shape, asset=asset)
```

This was missed on the last go-round.

gadomski (changes requested):

stactools_landsat/stactools/landsat/utils.py
carioca-au: Good Morning @gadomski,
     It wasn`t missed this part of the code is right. As I explained to you, ```item.ext.projection.set_transform``` and ```item.ext.projection.set_transform``` don't set anything in the Item, instead, it sets in the asset and because of that they MUST be inside of the loop.

stactools_landsat/stactools/landsat/utils.py
alexgleith: Hey @gadomski, the extension specifies that you can either set this at the root or per-asset.

We prefer it per-asset, and that's how the Element-84 S-2 COGs implemented it, see here: https://earth-search.aws.element84.com/v0/collections/sentinel-s2-l2a-cogs/items

So we do want to do this in the loop for each asset.

alexgleith:
Hey folks, we're really hoping to get this into production today. I know it's late in the US, but it would be awesome if we can get an approval or feedback asap! Sorry to hassle...

lossyrob (approved):
These changes LGTM

lossyrob:
Looks like there's formatting issues - you'll need to run `scripts/format` or `docker/format` and commit the results. Make sure that `docker/test` passes locally as it'll run the same script that CI does.

gadomski (approved):

gadomski:
Yeah sorry folks it looks like CI still isn't passing. My recommendations:
- Use `docker/cibuild` to do the same things that the Github actions do
- Install `yapf` in your IDE and use it to auto-do your formatting

PR #106, opened 2021-05-09
Re-indentation

carioca-au:
1. Re-indented **_item.ext.projection.set_transform_** and **_item.ext.projection.set_shape_**
2. Removed unused variable _**name**_

PR #107, opened 2021-05-10
Add option to eo instrument

carioca-au:
- USGS has changed their stac file, forcing us to add an additional test with a second option to eo:instrument depending on the informed type.


gadomski:
See https://github.com/stac-utils/stactools/pull/106#issuecomment-836633242 for instructions to rebase this pull request to current **master**.

PR #112, opened 2021-05-12
Use core's utilities in subpackages

gadomski:
A couple of the dataset subpackages had defined their own `call` and/or `cogify` functions which do more or less the same thing. This pull request adds `stactools.core.utils.convert.cogify` and updates subpackages to use core's `call` and `cogify` instead of their own. The core `cogify` function has sensible defaults (right now just `deflate` compression).

When implementing, I discovered that `cgls_lc100`'s `cogify` was broken due to a [dangling `'-co'` option](https://github.com/stac-utils/stactools/blob/5be2be0028eff7aee2d9f61accef760d5d676ee2/stactools_cgls_lc100/stactools/cgls_lc100/cog.py#L25). As part of the fix, I updated `cgls_lc100`'s test file to a subset of the original external file, since it was mondo and cogifying it was taking forever.

Also includes a small typo fix in README.


tests/utils/validate_cloud_optimized_geotiff.py
```diff
@@ -0,0 +1,449 @@
...
+#  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+#  OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#  THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#  DEALINGS IN THE SOFTWARE.
+# *****************************************************************************
+
+import os.path
```
lossyrob, line 32: Is there not a way to bring this in as a dependency rather than copy the code in?

lossyrob (approved):
LGTM, one question about the GDAL code but if copying in is the best route then :+1:

tests/utils/validate_cloud_optimized_geotiff.py
gadomski: I didn't find a pip package for it, the Proper Ecosystem Thing™ would probably be to PR https://github.com/rouault/cog_validator w/ pip support. But for now this was the best way I could think of.

PR #113, opened 2021-05-21
Pystac v1.0 compatability

gadomski:
Feedback appreciated, especially on the `StacIO` stuff, where I did my best to just port over the old behavior to The New Way™.

- [x] IO is broken until https://github.com/stac-utils/pystac/pull/348 lands in the next beta release
- ~[ ] Pystac 1.0 resolves parent collections, so some landset tests are broken because https://landsatlook.usgs.gov/data/collection02/landsat-c2l2-sr.json is not publicly accessible.~
- ~[ ] Other landsat tests break because landsat schema is no longer accessible: https://landsat.usgs.gov/stac/landsat-extension/schema.json.~
- [x] Aster and CGLS_LC100 both do not validate against the eo:bands schema. CGLS_LC100 has "Change-Confidence-layer" for a common name, and aster has 'yellow/green'.

cc @volaya since you said you might be working on this effort as well.


stactools_aster/stactools/aster/stac.py
```diff
@@ -121,30 +128,30 @@ def create_item(xml_href: str,
     item.common_metadata.instruments = [ASTER_INSTRUMENT]
 
     # eo
-    item.ext.enable('eo')
-    item.ext.eo.cloud_cover = xml_metadata.cloud_cover
+    eo = EOExtension.ext(item)
```
lossyrob, line 131: You'll have to add the extension to the item - the `.ext()` method doesn't automatically check to see if the extension is set. I had thought about this - having the `ext()` add the URI if it didn't already exist in the `stac_extensions` - but chose to avoid that check every time the method is invoked for performance, and also to be explicit about enabling it. 

You can do
```python
EOExtension.add_to(item)
eo = EOExtension.ext(item)
```

Also seeing how you expected it to work here makes me reflect on that API choice and whether it was the best. Any thoughts? cc @duckontheweb

gadomski: Yeah, I made a mistake here and didn't read the API completely, just looked at a couple tests (not to closely it seems!).

It can be hard to know in wheter `add_to` has been called previously in a given code path, so from a developer side I could see people just always using `add_to` then `ext` to protect themselves (e.g. if you're using an extension in a helper function). So I'd vote for `ext` to enable the extension if needed -- the check can't be that expense, right, just a dict lookup?

gadomski: Also very weird that none of the tests failed even though I never called `add_to` 🤔.

duckontheweb: Strictly speaking I think we should probably only treat a property as being part of an extension if the object explicitly implements that extension (i.e. has the URI in `stac_extensions`), so I think we should require developers to call `add_to` if they want to add the extension to the list. I'm worried that doing it behind the scenes in `ext` would mask a problem with an invalid object when people use PySTAC to read existing STAC objects, and it could also interfere with the odd case where someone is using their own custom fields that happen to overlap with an extension that they didn't intend to implement. I'm definitely open to pushback on any of these points, though.

I am thinking we should make 2 changes to PySTAC:

1) Raise an exception if the extension is not in `stac_extensions`. Right now it just returns any properties with the right name (e.g. `eo:cloud_cover`) whether the extension is actually implemented or not and it will also happily set those properties. I think @gadomski This is probably why it didn't error for you.
2) Ensure that `add_to` is idempotent. Right now it looks like it will append a URI as many times as the method is called. That way you could call it freely if you want to ensure that the extension is in `stac_extensions`

lossyrob: Good call, I think that's a good path forward. And good catch with add_to not being idempotent, that's a bug.

Perhaps we could also add a flag to `ext` to add and get the extension in a one liner, e.g.

```python
eo  = EOExtensions.ext(item, add_if_missing=True)
```

duckontheweb: > Perhaps we could also add a flag to `ext` to add and get the extension in a one liner, e.g.
> 
> ```python
> eo  = EOExtensions.ext(item, add_if_missing=True)
> ```

Yeah, I like that idea!

gadomski:
Ok, I've fixed up stuff to use extensions correctly. There's some issues with a couple subpackages, some of which are due to STAC v1 validation, some of which are landsat infrastructure issues. I've cataloged them as best as I can in this PR's description so we can try to axe them down.

For the aster+cgls_lc100, @lossyrob can you advise on what we should do about the band names?

The landsat stuff is trickier -- relying on someone's else's schemas and collections as a part of our test suite seems fragile.

lossyrob:
For Aster, it looks like the center wavelength falls in the "green" category from [this table](https://github.com/stac-extensions/eo#common-band-names), so I'd go with that....

For CGLS_LC100 I'd remove the common_name

gadomski:
Updated for the v0.2 reorganization (makes this PR a lot smaller).

gadomski:
@duckontheweb mind giving this one a once-over? It's much simpler now w/o the dataset packages.

src/stactools/core/io/__init__.py
```diff
@@ -12,22 +12,23 @@
...
 
 def read_text(href: str,
-              read_href_modifier: Optional[ReadHrefModifier]) -> str:
+              read_href_modifier: Optional[ReadHrefModifier] = None) -> str:
     if read_href_modifier is None:
-        return pystac.STAC_IO.read_text(href)
+        return StacIO.default().read_text(href)
     else:
-        return pystac.STAC_IO.read_text(read_href_modifier(href))
+        return StacIO.default().read_text(read_href_modifier(href))
```
duckontheweb, line 19: Since this is only used in `stactools.core.io.xml` maybe it would be simpler/more efficient to create an instance of `StacIO.default()` in a class attribute on `XmlElement` and then just reference it in `XmlElement.from_file`?

```python
class XmlElement:

    stac_io: pystac.StacIO = StacIO.default()

    ...

    @classmethod
    def from_file(
        cls,
        href: str,
        read_href_modifier: Optional[ReadHrefModifier] = None
    ) -> "XmlElement":
        if read_href_modifier is not None:
            href = read_href_modifier(href)
        text = cls.stac_io.read_text(href)
        return cls(etree.fromstring(bytes(text, encoding='utf-8')))

```

tests/cli/commands/test_cases.py
```diff
@@ -119,19 +120,19 @@ def test_case_3():
...
-                                       LabelClasses.create(
-                                           classes=['one', 'two'],
-                                           name='label')
-                                   ],
-                                   label_tasks=['classification'],
-                                   label_methods=['manual'],
-                                   label_overviews=overviews)
-        label_item.ext.label.add_source(image_item, assets=['ortho'])
+        LabelExtension.add_to(label_item)
+        label_extension = LabelExtension.ext(label_item)
```
duckontheweb, line 124: Just a note that you'll be able to replace this with

```python
label_extension = LabelExtension.ext(label_item, add_if_missing=True)
```

after then next release

tests/cli/commands/test_cases.py
```diff
@@ -119,19 +120,19 @@ def test_case_3():
...
+        label_extension.apply(label_description='ML Labels',
+                              label_type='vector',
```
duckontheweb, line 126: I think we're encouraging folks to use the [`LabelType.VECTOR` enum](https://github.com/stac-utils/pystac/blob/7f8e7aaf29c9882e717550b219228ccf27add74b/pystac/extensions/label.py#L48) whenever possible.

duckontheweb (commented):
Looks good to me, just a couple of minor comments!

src/stactools/core/io/__init__.py
lossyrob: Could be, though the utility might be useful for anyone else using the read_href_modifier pattern elsewhere in the future. The StacIO.default() call should return fast after the first as it provides a cached value

lossyrob (approved):

src/stactools/core/io/__init__.py
gadomski: Yup, at least one downstream package is currently using (https://github.com/stactools-packages/landsat/blob/main/src/stactools/landsat/ang_metadata.py#L4).

tests/cli/commands/test_cases.py
gadomski: I'll add a TODO.

duckontheweb (approved):
👍

PR #121, opened 2021-06-02
Offset boundaries from center of pixel for landsat

### LINKED ISSUE #117: landsat8-c2-l2 proj:bbox is off by a half-pixel
Hi all, looking for some debugging help here.

We have an item, https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-8-c2-l2/items/LC08_L2SP_231060_20200112_20200823_02_T1, where the `proj:bbox` doesn't match what's given by rasterio:

```python
In [1]: import pystac, planetary_computer, rasterio

In [2]: item = pystac.Item.from_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-8-c2-l2/items/LC08_L2SP_231060_20200112_20200823_02_T1")

In [3]: ds = rasterio.open(planetary_computer.sign(item.assets["SR_B3"].href))

In [4]: item.properties["proj:bbox"]
Out[4]: [60900.0, -116400.0, 288900.0, 115800.0]

In [5]: list(ds.bounds)
Out[5]: [60885.0, -116415.0, 288915.0, 115815.0]
```

The bounds seem to be off by 15, half a pixel.

---

https://gist.github.com/TomAugspurger/cf5cdc4d6e94ec3757411dadae6e8558 has the XML MTL file for this item. IIUC, we get `proj:bbox` at https://github.com/stac-utils/stactools/blob/0e7509b9596379612cf29713ee111ed673d68cec/stactools_landsat/stactools/landsat/mtl_metadata.py#L96-L120. Looking at the file, we do indeed have `60900.000` for the left-most bound. Is this pretty clearly a case of the metadata file being wrong, which should be reported to USGS? Or might rasterio / GDAL be doing something wrong? Or might stactool's derivation of `proj:bbox` failing to take something into account? Any help would be appreciated.

gadomski:
Fixes #117.

We use the SR ground sample distance as the bonding box pixel resolution. We do our best to be intelligent about the lat+lon offsets to account for assets near the poles.

### DIFF
--- a/stactools_landsat/stactools/landsat/mtl_metadata.py
+++ b/stactools_landsat/stactools/landsat/mtl_metadata.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Any, Dict, List, Optional
 
 from pystac.utils import str_to_datetime
+from pyproj import Geod
 
 from stactools.core.utils import map_opt
 from stactools.core.projection import transform_from_bbox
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks for handling this!

lossyrob (approved):

PR #122, opened 2021-06-06
Fix issue in sentinel 2 metadata where 0 z coordinates are listed.

lossyrob:
There are some product metadata files that contain 0 z coordinate values in the product metadata that is used to construct STAC Item geometries. This causes the Item geometry to be invalid. This change modifies the parsing logic to filter out '0' string values in the coordinates to avoid this issue.


gadomski (approved):
Looks good as is, I have three questions mostly for my own education:
- Is there chance of a boundary granule footprint that includes `0` as a x or y value that could get caught by the added check?
- Are 3D geometries useful? Could this be solved by only supporting 2D?
- Does one of the test files include the bogus z value?

lossyrob:
> Is there chance of a boundary granule footprint that includes 0 as a x or y value that could get caught by the added check?

Looking at the precision of the other x/y values, I don't believe so, but could not guarantee it. I found it difficult to find documentation on this part of the sentinel product metadata, so there's some assumptions made here.

> Are 3D geometries useful? Could this be solved by only supporting 2D?

The example I found had all 0 values for z, which is common in a bug I've seen elsewhere where the default z values are printed out unintentionally. I have a hunch that a similar issue happened here with some scenes printing the default. I couldn't find documentation of meaningful Z footprint values.

> Does one of the test files include the bogus z value?

The added test case reads an example product metadata file that was found to have the bogus Z values.

gadomski:
Roger, that all sounds good to me. Once CI passes I can merge. Thanks!

PR #135, opened 2021-06-11
Update external data for separate repos

cuttlefish:
I'm not sure if this is a good way to handle it or not, but at least it moves the definition of the data to the dataset module rather than the main library.

Any thoughts @gadomski ?

### DIFF
--- a/src/stactools/testing/test_data.py
+++ b/src/stactools/testing/test_data.py
@@ -5,19 +5,20 @@ from zipfile import ZipFile
 
 import requests
 
-EXTERNAL_DATA = {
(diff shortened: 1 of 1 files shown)


cuttlefish:
The caller would do something like:
`from stactools.testing import TestData

_external_data = {
    'AST_L1T_00305032000040446_20150409135350_78838.hdf': {
        'url':
        ('https://ai4epublictestdata.blob.core.windows.net/'
         'stactools/aster/AST_L1T_00305032000040446_20150409135350_78838.zip'),
        'compress':
        'zip'
    }
}

test_data = TestData(__file__, _external_data)
`

gadomski (approved):
Looks good to me.

PR #161, opened 2021-07-15
Move issue and PR templates

jbants:
The templates are sitting in the wrong location int he .github directory.


gadomski (approved):
Right on, shoulda caught that in my review. Thanks for the fix.

PR #162, opened 2021-07-16
Push docker images to GitHub Container Registry via GitHub Action

cuttlefish:
Push Docker images to GitHub Container Registry via the CI Actions

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -102,9 +102,11 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: ${{ env.DOCKER_REGISTRY }}/${{ env.DOCKER_IMAGE_NAME }}
+          flavor: |
+            suffix=-dev
           tags: |
-            type=ref,event=branch,suffix=-dev
-            type=ref,event=pr,suffix=-dev
+            type=ref,event=branch
+            type=ref,event=pr
       -
         name: Build and push dev
         uses: docker/build-push-action@v2
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -92,7 +92,8 @@ jobs:
           context: .
           file: docker/Dockerfile
           target: main
-          push: true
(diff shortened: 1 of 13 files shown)


lossyrob:
This looks great - so users/downstream projects will be able to use `ghcr.io/stac-utils/stactools` as a public image?

Should we include some README docs around using the published image as part of this PR?

cuttlefish:
@lossyrob I was planning on holding off to ensure that the release images are working properly before updating the README. Would it be possible to merge this PR and do a 0.2.1a2 release to verify this?
I'll also have to update the `./docker/` scripts in the `stactools-packages` repos to use the new images.
I was also thinking about adding a `pull` script for people that want to use the scripts, but don't want to do full builds.
If that sounds good, I'll make issues for those and should be able to get them done quickly.

cuttlefish:
The docker scripts have been updated to use the new image registry. I've added a brief note in the README.
`docker/build` now takes optional `--pull` and `--no-cache` flags.
There's now a `docker/pull` for those that want to run the `docker/` scripts but who don't want to run a build.

lossyrob (approved):

lossyrob:
Sounds good, let's get this in and see how it works and make changes from there.

I'm a little unfamiliar with the pattern of pulling docker images for dev runs, and building locally against the fully qualified name (e.g. `ghcr.io/stac-utils/stactools:latest`). I'm used to the local dev environment having to build the containers itself with an unqualified name (e.g. just `stactools`), and then the scripts run those containers - at the time of push, those images are re-tagged as the fully qualified versions and then pushed. Thinking through it a bit, I don't see any disadvantage to just pulling the image and running the dev container without having to build it, and the building to match the qualified name locally - maybe that there's less of a distinction between a pulled image or a locally modified one? Either way, just pointing out the difference in process that I'll be looking at using this method. Looks solid though, thanks for sorting this out.

cuttlefish:
In the past I've found little difference in whether fully qualified names are used for the dev images or not. But I was second guessing myself last night... If we run into any problems we can easily change it. We could still pull images and retag them to the local name if desired.

PR #173, opened 2021-07-20
Update to version 0.2.1a2

cuttlefish:
Also bump pystac to 1.0.0
It drops support for Python 3.6, so dropping that.


gadomski (approved):
LGTM. I'm thinking we should probably start pushing for an v0.2.1rc1 after this to get a stable release pinned to PySTAC v1 as soon as we can.

gadomski (changes requested):
We'll need to update the CHANGELOG as well.

gadomski (approved):

gadomski (approved):

Issue #165, opened 2021-07-18
Should Jupyter be a dev dependency?


cuttlefish:
There's an existing `docker/notebook` script to run a Jupyter notebook. I thought there was a `script/notebook` to match it, but there isn't.
The `docker/notebook` script doesn't work because Jupyter isn't listed in the dev dependencies.

Should we:
1. Remove `docker/notebook`
2. Add Jupyter as a dev dependency and add `script/notebook`
3. Create a separate docker image that includes Jupyter and use that in `docker/notebook`

There's some work going into the `stactools-packages` repos to add example Jupyter notebooks, so I don't really like the idea of removing it entirely.
I think 3 is likely adding unneeded complexity and also doesn't make it easy for non-Docker users to run notebooks.
I'm leaning towards option two but am open to suggestions (or alternatives I've missed).

gadomski:
> Add Jupyter as a dev dependency and add script/notebook

This makes sense to me, 👍🏽.

PR #174, opened 2021-07-20
Add script for running Jupyter notebooks

cuttlefish:

**Related Issue(s):**
#165 

**Description:**
Add a script for running Jupyter notebooks. There was already a docker script to do so.
Add Jupyter as a dev dependency.

**PR checklist:**

### DIFF
--- a/scripts/notebook
+++ b/scripts/notebook
@@ -19,7 +19,6 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     else
         jupyter notebook \
             --ip=0.0.0.0 \
-            --port=8888 \
-            --allow-root # Useful if run within the Docker containers
+            --port=8888
     fi
 fi
--- a/scripts/notebook
+++ b/scripts/notebook
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -e
+
+if [[ -n "${CI}" ]]; then
+    set -x
+fi
+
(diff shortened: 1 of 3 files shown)


scripts/notebook
```diff
@@ -0,0 +1,25 @@
...
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        jupyter notebook \
+            --ip=0.0.0.0 \
+            --port=8888 \
+            --allow-root # Useful if run within the Docker containers
```
gadomski, line 23: It's probably very minor, but can we add an argument to the script to opt-in to `--allow-root`, and use that flag in `scripts/docker`? I don't know enough to know that this is a security hole, but an implicit `--alow-root` smells funny. Feel free to tell me I'm wrong and this isn't an issue at all :-D.

gadomski (changes requested):

scripts/notebook
cuttlefish: Yeah, fair enough. I'm not enthralled that the Docker containers are running as root, but it can result in wonky permissions problems when mounting local volumes.
The use-case I was considering for running in the Docker containers doesn't actually make sense here, so I'll just remove it.

gadomski (approved):

PR #178, opened 2021-07-23
Use fsspec for fetching external testing files

gadomski:
**Related Issue(s):** n/a

**Description:** This enables external files stored on all of the systems supported by fsspec.

Also includes:
- Two `extra_requires` options. `stactools[s3]` installs `s3fs`, which is required to use fsspec on s3 hrefs. `stactools[all]` includes `s3` and any future extra_requires.
- CI tests for `stactools[all]`, to make sure we are testing both with and without `extra_requires`.
- Version bump for `fssepc`.
- Unit tests for `stactools.testing`, which leads to saying "test" a lot :-).
- README instructions for install w/ optional dependencies.

**PR checklist:**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,8 +40,10 @@ install_requires =
     lxml ~= 4.6
 
 [options.extras_require]
-all = s3
-s3 = s3fs ~= 2021.7
+all =
+    %(s3)s
+s3 =
+    s3fs ~= 2021.7
 
 [options.packages.find]
 where = src
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,6 @@ install_requires =
     lxml ~= 4.6
 
 [options.extras_require]
-all = s3
 s3 = s3fs ~= 2021.7
 
 [options.packages.find]
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,12 +32,16 @@ install_requires =
     pystac[validation] ~= 1.0.0
     aiohttp ~= 3.7
     click ~= 7.1
-    fsspec ~= 2021.6.0
+    fsspec ~= 2021.7
     requests ~= 2.25
     Shapely ~= 1.7
     pyproj ~= 3.0
     rasterio ~= 1.2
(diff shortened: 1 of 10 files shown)


gadomski:
Converting to draft because I'm going to add some kwargs to the external data structure, as I need to use an endpoint url as documented here: https://s3fs.readthedocs.io/en/latest/#self-hosted-s3.

cuttlefish:
Nice! This will be really useful.
I won't hold up this PR, but I'll add it to the Docker images.
`pip install stactools[all]` ended up not working, right?

cuttlefish:
@gadomski Sorry for messing with your branch, but I updated the Dockerfile so that I could test this. I also got the `all` `extras_require` working, although I'm not sure how robust it is.

cuttlefish (approved):
Looks good!

setup.cfg
```diff
@@ -32,13 +32,19 @@ install_requires =
...
     pyproj ~= 3.0
     rasterio ~= 1.2
     lxml ~= 4.6
 
+[options.extras_require]
+all =
+    %(s3)s
+s3 =
+    s3fs ~= 2021.7
+
```
cuttlefish, line 47: @gadomski Are you OK with this?

gadomski: Can you explain what it's doing? Is it expanding the `s3` variable to its string value? Then -- maybe? I guess I'm not convinced its worth it to do backflips to enable an `setuptools[all]` quite yet. My impression is most users using non-standard features (e.g. if we enabled Azure Blob storage URLs or something like that) would be working a bit off the beaten path.

cuttlefish: Yeah, I'm not enthralled with the level of non-standard cleverness. But if we end up adding more it might be useful.
It's using `configparser`'s [BasicInterpolation](https://docs.python.org/3/library/configparser.html#configparser.BasicInterpolation).

gadomski: Ok, I'm good w/ keeping it. Merge away! Thanks for the updates!

PR #182, opened 2021-07-27
Add `mypy`

### LINKED ISSUE #155: Enable mypy
Static typing is good we and we should use it if we can.

cuttlefish:
**Related Issue(s):**
Closes #155 

**Description:**
Adds `mypy` and hooks it into `scripts/lint`.
This will be much more useful once stac-utils/pystac#579 is merged and released.
The settings are currently fairly permissive.

**PR checklist:**

- ~[ ] Documentation has been updated to reflect changes, if applicable.~

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -23,6 +23,13 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         yapf -dpr ${DIRS_TO_CHECK[@]}
         # Lint
         flake8 ${DIRS_TO_CHECK[@]}
+        # Type checking
+        for module in src/*/; do
+            if [[ "$module" == *.egg-info/ ]] ; then continue; fi
+            for submodule in "$module"/*; do
+                mypy -p "$(basename "$module").$(basename "$submodule")"
+            done
+        done
 
     fi
 fi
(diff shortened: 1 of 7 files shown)


scripts/lint
```diff
@@ -23,6 +23,13 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         yapf -dpr ${DIRS_TO_CHECK[@]}
         # Lint
         flake8 ${DIRS_TO_CHECK[@]}
+        # Type checking
+        for module in src/*/; do
+            if [[ "$module" == *.egg-info/ ]] ; then continue; fi
+            for submodule in "$module"/*; do
+                mypy -p "$(basename "$module").$(basename "$submodule")"
```
gadomski, line 30: I've found in the past that errors can slip through if `--namespace-packages` isn't included.

gadomski (changes requested):
Just one change to the `mypy` invocation, otherwise looks good. We should probably open another issue to add type signatures to all `stactools.{core|cli|testing}` functions/methods as well.

scripts/lint
cuttlefish: Yeah, it's in `mypy.ini` (which we need for the module ignore list). `mypy` automatically picks it up by location. There would be a ton of "missing module" errors otherwise.

gadomski (approved):

PR #189, opened 2021-08-01
Add type annotations

### LINKED ISSUE #183: Populate type signatures
Ensure that the signatures to all 'stactools.{core|cli|testing}' functions/methods have type annotations.

See #155 and #182

cuttlefish:
Closes #183 
`pystac 1.1.0` is required because it now publishes type information in its packages.

I had to add some error handling code to appease `mypy`, so I'd like to get this looked at carefully. I modified a lot of code I haven't dealt with before.

**PR checklist:**

### DIFF
--- a/src/stactools/core/io/__init__.py
+++ b/src/stactools/core/io/__init__.py
@@ -22,13 +22,13 @@ def read_text(href: str,
 class FsspecStacIO(DefaultStacIO):
     def read_text_from_href(self, href: str, *args: Any, **kwargs: Any) -> str:
         with fsspec.open(href, "r") as f:
-            return f.read()
+            return str(f.read())
 
     def write_text_from_href(self, href: str, txt: str, *args: Any,
                              **kwargs: Any) -> None:
         with fsspec.open(href, "w") as destination:
-            return destination.write(txt)
+            destination.write(txt)
 
 
-def use_fsspec():
+def use_fsspec() -> None:
     StacIO.set_default(FsspecStacIO)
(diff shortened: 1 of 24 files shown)


src/stactools/core/io/__init__.py
```diff
@@ -22,13 +22,13 @@ def read_text(href: str,
 class FsspecStacIO(DefaultStacIO):
     def read_text_from_href(self, href: str, *args: Any, **kwargs: Any) -> str:
         with fsspec.open(href, "r") as f:
-            return f.read()
+            return str(f.read())
```
gadomski, line 25:
```suggestion
            return str(f.read(), encoding='utf-8')
```

Ref: https://docs.python.org/3/library/stdtypes.html#str

src/stactools/core/io/xml.py
```diff
@@ -52,8 +52,8 @@ def find_attr(self, attr: str, xpath: str) -> Optional[str]:
         return None if node is None else node.get_attr(attr)
 
     @property
-    def text(self):
-        return self.element.text
+    def text(self) -> Optional[str]:
+        return str(self.element.text)
```
gadomski, line 56:
```suggestion
        if isinstance(self.element.text, bytes):
            return str(self.element.text, encoding='utf-8')
        elif isinstance(self.element.text, str):
            return str(self.element.text)
        else:
            assert (self.element.text == None)
            return None
```

gadomski (changes requested):

src/stactools/core/io/__init__.py
cuttlefish: Forcing the UTF8 decoding was causing tests to fail, so I did something similar to the XML handling change you suggested below.

cuttlefish:
Thanks @gadomski! I wasn't sure how to handle the statically typed file inputs, but your suggestions look good.

gadomski (approved):

PR #190, opened 2021-08-02
Install stactools as part of the readthedocs build

gadomski:
**Related Issue(s):** #143 

**Description:** As viewable [here](https://readthedocs.org/projects/stactools/builds/14371316/), our readthedocs build is failing because there's no `stactools` in the path. I'm guessing this is because we moved the code into a `src/` subdirectory as a part of the v0.2 update. By installing stactools into the virtualenvironment, hopefully this will go away.

This is hard to test b/c docs for pull requests isn't enabled (I think). @lossyrob I think you own our RTD account, would it be possible to turn on PR docs per https://docs.readthedocs.io/en/stable/pull-requests.html so we can see if we're fixing things?

**PR checklist:**

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,4 +16,6 @@ formats:
 python:
   version: 3.8
   install:
+    - method: pip
+      path: .
     - requirements: requirements-dev.txt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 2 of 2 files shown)


cuttlefish:
:+1:  for enabling PR docs builds (or making them public, or whatever else we need to verify this)

cuttlefish:
Should we add a step to the CI for running the docs build?

gadomski:
It _looks_ like there's a build status report as part of the PR integration from ReadTheDocs (https://docs.readthedocs.io/en/stable/pull-requests.html#features). If so, I _think_ that should be a good enough check on our docs build?

cuttlefish (approved):

Issue #70, opened 2021-03-07
Add linting various use-cases


TomAugspurger:
Over in stac-vrt, we rely on the presence of some extensions to efficiently build VRTs from STAC metadata. This issue is a request for two things

1. Write a tool (CLI) to check that a STAC item contains the extended data needed by stac-vrt.
2. Verify that the STAC items generated by `stactools` include the extended data (as a linting check, that can optionally be skipped when it's not applicable / desired for whatever reason).

https://stac-vrt.readthedocs.io/en/latest/#building-stac-vrt-compatible-stac-items has the list of items currently required by stac-vrt.

gadomski:
We have `stac validate` now, so that would be a nice place to add in warnings e.g. these two.

gadomski:
Going to close this as OBE due to [stac-validator](https://github.com/stac-utils/stac-validator) and [stac-check](https://github.com/stac-utils/stac-check).

PR #191, opened 2021-08-03
PoC: Add an optional validator to verify required fields for `stac-vrt`

cuttlefish:
Addresses a request from #70 

This is a proof of concept. I'm not enthralled with the implementation but would like some feedback on the general viability of it.

A `RequiredFieldsValidator` class is added that checks whether a specified (by `STACObjectType`) list of fields exists in the `STACObject`. I thought this might be useful for cases other than `stac-vrt`. The normal JSON Schema validator is run, then any other enabled validators are run. The errors are combined and displayed as before.
`pystac` has the validator set gobally, so I use a context manager to safely change the setting and restore it.

A possible alternative implementation could insert JSON Schemas into the existing validator. My JSON Schema is rusty so the above was more expedient. I thought it was also interesting as a stub to allow people to write more complex validators.

In general, would it make sense to move the main logic for validation to `core` so that it's more easily accessible programmatically? Perhaps it might be of value to plugin authors?

I was debating whether this should be implemented in a separate package. Probably? It might be useful to include the `RequiredFieldsValidator` in the main `stactools` package.

**PR checklist:**


gadomski:
I like the idea! I didn't even think of using `STACValidator`. The only limitation I see right now is that it doesn't have a "warning" concept, which is something that I think would be useful from the command-line side; e.g. the `stac-vrt` recommended fields could be a warning that's on by default, since we want to nudge people in that direction. I'll respond inline to a couple main questions, then add some review comments as well.

> A possible alternative implementation could insert JSON Schemas into the existing validator. 

I like your current approach.

> In general, would it make sense to move the main logic for validation to core so that it's more easily accessible programmatically? Perhaps it might be of value to plugin authors?

Yup, I think the validation should absolutely move to `core`. This is in line w/ my intention when making `stac validate`, I just didn't get around to it :-). 

> I was debating whether this should be implemented in a separate package. Probably?

My vote is no unless it grows really big. We require `pystac[validation]` so we aren't adding any dependencies by keeping it here, and it's going to be generally useful.

src/stactools/cli/commands/validate.py
```diff
@@ -15,25 +17,42 @@ def create_validate_command(cli):
...
 
         Prints any validation errors to stdout.
         """
         object = pystac.read_file(href)
 
-        if isinstance(object, Item):
-            errors = validate(object, None, False, links, assets)
-        else:
-            errors = validate(object, object, recurse, links, assets)
+        validators = [None]  # Start with the existing default
```
gadomski, line 40: I think we should convert all of our checks to validators to keep things consistent, i.e. `LinksValidator` and `AssetsValidator`.

src/stactools/cli/commands/validate.py
```diff
@@ -15,25 +17,42 @@ def create_validate_command(cli):
...
+                }))
+
+        errors = []
+        for validator in validators:
+            if isinstance(object, Item):
+                errors += root_validate(validator, object, None, False, links,
+                                        assets)
+            else:
+                errors += root_validate(validator, object, object, recurse,
+                                        links, assets)
```
gadomski, line 55:
```suggestion
        validators.insert(0, None)
        for validator in validators:
            pystac.validators.set_validator(validator)
            if isinstance(object, Item):
                errors += validate(object, None, False, links, assets)
            else:
                errors += validate(object, object, recurse, links, assets)
        pystac.validators.set_validator(None)
```

I don't know if we need the context manager, since the "default" validator activated by setting it to `None`.

src/stactools/cli/commands/validate.py
```diff
@@ -15,25 +17,42 @@ def create_validate_command(cli):
...
-            errors = validate(object, object, recurse, links, assets)
+        validators = [None]  # Start with the existing default
+        if stac_vrt:
+            validators.append(
+                RequiredFieldsValidator({
+                    STACObjectType.ITEM:
+                    ["proj:bbox", "proj:epsg", "proj:shape", "proj:transform"]
+                }))
+
+        errors = []
```
gadomski, line 48: One thing I've been thinking about for validation is that we should use an object rather than a simple string. It can start simple, e.g.

```python
@dataclass
class ValidationResult:
    message: str


class Error(ValidationResult):
    def echo(self):
        click.secho("ERROR", fg="red", nl=False)
        click.echo(f" {self.message}")


class Warning(ValidationResult):
    def echo(self):
        click.secho("WARNING", fg="yellow", nl=False)
        click.echo(f" {self.message}")
```

Or something like that. We could maybe even return these results from our custom validators, since the validators can both raise validation errors and return values?

src/stactools/cli/commands/validate.py
```diff
@@ -47,6 +66,28 @@ def validate_command(href, recurse, links, assets):
...
+        finally:
+            RegisteredValidator.set_validator(old_validator)
+
+    if validator:
+        with modded(validator):
+            return validate(object, root, recurse, links, assets)
+    else:
+        return validate(object, root, recurse, links, assets)
+
+
```
gadomski, line 90: See my above comment about the context manager, but I'm not sure we need this.

gadomski (changes requested):

src/stactools/cli/commands/validate.py
```diff
@@ -15,25 +17,42 @@ def create_validate_command(cli):
...
+                        "(only useful for Catalogs and Collections)"))
     @click.option("--links/--no-links",
                   default=True,
                   help=("If false, do not check any of the objects's links."))
     @click.option(
         "--assets/--no-assets",
         default=True,
         help=("If false, do not check any of the collection's/item's assets."))
-    def validate_command(href, recurse, links, assets):
+    @click.option("--stac-vrt",
```
TomAugspurger, line 28: It'd be nice to generalize this, since these kinds of checks will be useful to several projects (GDAL dev now includes a driver for STAC items, so the `stac-vrt` library is basically obsolete now).

I'm not sure what name would be best, but perhaps just `--proj`?

src/stactools/cli/commands/validate.py
```diff
@@ -47,6 +66,28 @@ def validate_command(href, recurse, links, assets):
...
+        try:
+            yield
+        finally:
+            RegisteredValidator.set_validator(old_validator)
+
+    if validator:
+        with modded(validator):
+            return validate(object, root, recurse, links, assets)
+    else:
+        return validate(object, root, recurse, links, assets)
```
TomAugspurger, line 88: As Pete mentioned we might not need a contextmanager, but as an FYI `contextlib.nullcontext` can be useful for removing the need for the `if / else`  in cases where you maybe need a contextmanager: https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext.

src/stactools/cli/commands/validate.py
cuttlefish: I can't think of anything better. It's a bit misleading because the `proj` extension only required the `proj:epsg` field. But it will be better than `--stac-vrt`.

src/stactools/cli/commands/validate.py
cuttlefish: I was using the context manager for two reasons. One was to ensure that the validator would be set to the pre-existing setting, even if an uncaught exception stopped our command from completing. With the setting being global, any other use of `object.validate()` will use whatever validator threw the exception. [I think that at the moment this won't happen because we catch all exceptions]
The other was to allow users to change the main validator. That isn't really an issue at the moment, but if people are using `stactools` programmatically it might cause grief, since running our `validate` will change the behaviour of future `object.validate()`s.
Another option would be to add a `validate(self, validator)` to `STACObject` in `pystac` which would remove the need for this.

gadomski: Ah gotcha, ok the context manager makes sense then.

>  Another option would be to add a validate(self, validator) to STACObject in pystac which would remove the need for this.

I agree with this, much prefer declarative instead of global-ish magic :-).

src/stactools/cli/commands/validate.py
gadomski: Would it make sense to have an enumeration of `--best-practices` and then have a bunch of named best practices validators? E.g. `proj`, `eo`, `summaries`, etc, with a default `all`? So you'd call `stac validate --best-practices proj,summaries catalog.json` or whatever.

gadomski:
Closing as Overcome By Events.

Issue #179, opened 2021-07-23
`stac create-item` and associated API function


gadomski:
**Describe the solution you'd like**
I think there's some boilerplate in "open a file in rasterio and create an Item from it". E.g. https://github.com/gadomski/alos-dem/blob/d4bbd4a7ed6d2d362626a829dff73fa1e6145d87/src/stactools/alos_dem/stac.py#L12-L32. It'd be nice to call `stactools.core.create_item(href)` and get back something as simple as possible. That'd include:
- geometry and bbox populated
- id from filename
- maybe the proj extension implemented if the thing's not in EPSG:4326
- asset w/ the original href

This would simplify standing up new dataset packages, and would be an equivalent to `rio info` for `stac`. Could even make the command called `stac info`?

**Describe alternatives you've considered**
Doing it myself all the time for lots of new datasets :-).

**Additional context**
I thought I remembered there being an issue for this already, but I couldn't find it after some quick Friday-afternoon searching. If there is feel free to close.

cholmes:
I like this idea. Perhaps your confusion on an issue may be that there already is a command called `stac info`:

```
% stac info https://www.planet.com/data/stac/catalog.json
Catalog ID: planet
------------------
   CATALOGS: 3 
COLLECTIONS: 2 
      ITEMS: 9 (extensions: https://stac-extensions.github.io/projection/v1.0.0/schema.json,https://stac-extensions.github.io/view/v1.0.0/schema.json,https://stac-extensions.github.io/eo/v1.0.0/schema.json)
```

Though sounds like yours is to work at the item level? That'd be cool for sure.

Issue #196, opened 2021-08-30
Add API+CLI ability to add the `raster` extension (and maybe other extensions) to an item


gadomski:
**Is your feature request related to a problem? Please describe.**
Many useful `Item` attributes, especially extension attributes, are trivial to construct by (e.g.) reading the metadata of a raster file.

**Describe the solution you'd like**
We should provide a function, available via both the API and the CLI, to add common extensions to an item by reading the metadata from an asset.

volaya:
@gadomski I want to start working on this, but I was thinking that there should be some module in pystac to handle the raster extension stuff (like we have for other extensions, under `pystac/extensions`. I haven't seen any  issue open in there about this. Should I put all the logic for the raster extension in stactools and then we discuss about moving part of it to pystac if that's better?

volaya:
Nevermind, I had an older version of pystac, and I have just seen that pystac already has the raster extension support. Sorry for the noise

cholmes:
This seems similar to #159. I'll close that one.

kylebarron:
Note there might be some overlap between this and rio-stac: https://github.com/developmentseed/rio-stac/, which adds the `raster` extension.

cholmes:
Thanks for mentioning that @kylebarron. I think the user flows are a bit different - I suspect this will mostly be used to take an existing STAC collection or item and add the raster info to each item, while rio stac lets you create a new item with raster.

But potentially may be able to share some code between the two.

PR #201, opened 2021-09-13
Add stactools.core.create.item

gadomski:
**Related Issue(s):** #179 

**Description:** This creates an item from a `rasterio` openable asset. The function tries to fill in as much data as possible from the `rasterio` dataset, but generally created items should be enhanced after creation with additional properties and STAC extensions.

Also includes a drop-dead simple command line interface.

**PR checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- `stactools.core.create.item` and associated CLI subcommand ([#201](https://github.com/stac-utils/stactools/pull/201))
+
+### Fixed
+
+- Typing for Python 3.7 in `stactools.core.projection` ([#201](https://github.com/stac-utils/stactools/pull/201))
+
 ## stactools 0.2.2
 
 ### Added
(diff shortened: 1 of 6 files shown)


gadomski:
In https://github.com/stac-utils/stactools/issues/196#issuecomment-918656480 @kylebarron brought my attention to `rio stac`, which does more-or-less the same thing as this PR (with a slightly different approach). My initial instinct is to close this PR and recommend users favors `rio stac`, as it is (a) more feature-ful and (b) already extant, but I'd be curious whether @lossyrob, @cholmes, or others think it's worth having a "make an Item from an asset" functionality over here in **stactools** as well.

If this PR _is_ closed, 7d743983d4df144ca5056401605e18d9a689dec7 should be pulled out to its own PR.

cholmes:
I'll let @lossyrob speak from the API perspective, but from a CLI perspective I think it makes good sense to include a 'create-item' command in the `stac` cli. If behind the scenes that uses `rio stac` directly I think that's fine / perhaps preferable for less code duplication. But I see stactools as the 'first' command line tool someone coming to STAC would use, and they might not know to install `rio stac`. But then there are also lots of rasterio users who may learn about stac from `rio stac`. 

If it simplifies the code I could see `stac create-item` being really simple, but then also in its help / docs it'd point users at `rio stac` if they want more options.

lossyrob:
> If it simplifies the code I could see stac create-item being really simple, but then also in its help / docs it'd point users at rio stac if they want more options.

This seems like a good approach to me.

lossyrob (approved):

PR #204, opened 2021-09-14
Added addraster command

volaya:
First version of functionality to add raster extension content to items. Just to get some feedback on the approach, and if everything is OK, I will add unit tests

Note: I am using GDAL instead of rasterio (used in other parts of the code and AFIAK, preferred over plain GDAL), since there seems to be no rasterio method to compute stats and histograms, other than getting the layer band and then relying on numpy methods. This is fine for smaller layers, but  not for larger ones. That would require windowed reading and a more complex logic to be implement in the stactools code, which would also be less performant. 

Another note: For the `spatial_resolution` property I am taking the X size of the pixel, obtained for the GDAL geotransform. The  spec says that it is the "average" pixel size. I don't know if I should compute the average of the X and Y sizes, or how to corectly set that field.

**Related Issue(s):**

https://github.com/stac-utils/stactools/issues/196


src/stactools/core/addraster.py
```diff
@@ -0,0 +1,49 @@
+import logging
+
+from osgeo import gdal
```
gadomski, line 3: I don't believe `gdal` is an explicit dependency in https://github.com/stac-utils/stactools/blob/7aa3ae09e04aa7ebb48ac24da649979d32367f34/setup.cfg#L33-L42, so it should be added there.

volaya: does it have to be explicit? it is a requirement for rasterio, so if rasterio is installed, gdal should be as well.

gadomski: It does. GDAL's system libraries are a requirement for rasterio, but not its Python packages. E.g. this is from my personal stactools virtualenv:

```
$ python -c 'import stactools.core; print(stactools.core.__version__)'
0.2.2
$ python -c 'from osgeo import gdal'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'osgeo'
```

volaya: Right, good catch. I will add it

cuttlefish:
@volaya Hey Victor, long time no see 😄 
I think the addition of GDAL as a dependency means we'll need some changes in the Conda and Docker config to get the CI to pass. Maybe it'd be easiest if you granted me write on your fork so I can push to this branch? Or I can make my own branch and you can pull in the changes.

volaya:
@cuttlefish I have added you as collaborator on my fork. Let me know if I can do anything else

Thanks!

cholmes:
Any word on this @cuttlefish? I'd love to get this one in.

cholmes:
Oh, also it'd be really awesome to have this work on a catalog or collection, iterating over every item and adding raster fileds to the item if it doesn't yet have them. I could see `addraster` being able to work on a catalog / collection. Or a nice alternative could be to add it to the `copy` command as an option.

cuttlefish:
Hey, sorry about that. It should be up-to-date now. @volaya can you sanity check that I didn't bungle the conflicting merge from `main`?

volaya:
Looks good to me. Everything in place

gadomski (approved):

PR #209, opened 2021-11-12
Update to `pystac` version 1.2

cuttlefish:
`pystac` 1.2 has been released. This PR updates to that version.
It also matches the `mypy` configuration by switching to `strict`. This required adding explicit reexports to the core module and updating `click` to version 8. It includes more thorough type annotations for `click` than were previously available and resolves issues involving untyped decorators. Any `stactools-packages` or other projects importing `stactools` should remove `types-click` from their builds.

**PR checklist:**

- [ ] ~Documentation has been updated to reflect changes, if applicable.~


cuttlefish:
Thanks @gadomski, I just checked and don't think there's anything that needs updating. We'll just have to make sure to remove `types-click` from the `requirements-dev.txt` in the `stactools-packages` repos when we update them to the next version of `stactools`. I'll make a PR for the template repo once we get closer.
Most of the `stactools-packages` should be pinned to a specific version of `stactools`, so they can be done piecemeal.

Issue #216, opened 2022-01-12
Remove GDAL requirement


gadomski:
As discussed in #215, GDAL and rasterio don't always play nicely together. Everything that we added in #204 could be done w/ rasterio, so we should move everything over to that and remove the GDAL dependency.

sgillies:
@gadomski I can take this one on during my current sprint at work. It's highly related to https://github.com/rasterio/rasterio/issues/244#issuecomment-928542575. I think my approach will be to copy the rio-calc chunker in here to provide efficient statistics computation.

Additionally https://github.com/stac-utils/stactools/blob/0aa483dd7b5b9891cab3839d12c7df4d9037703d/src/stactools/core/utils/convert.py#L9 could be rewritten to not use gdal_translate. Without requiring conda it's not possible for a Python package to specify that it depends on a program in a non-Python package. Maybe that's a different issue?

gadomski:
> Without requiring conda it's not possible for a Python package to specify that it depends on a program in a non-Python package. Maybe that's a different issue?

Yeah, I guess we've been implicitly relying on being in a geo-enabled box so GDAL stuff is available, but I agree that removing that implicit dependency (and the command-line callout) would be preferable.

geospatial-jeff:
Could rewrite `cogify` to use https://github.com/cogeotiff/rio-cogeo

gadomski:
@geospatial-jeff we really aren't doing anything special in the `cogify` function (other than defaulting to `deflate` compression). Might make sense to remove it all together and just use **rio-cogeo** in any downstream packages that need to cogify? I'm not sure what the value of having a `cogify` function in **stactools** is, if there's a Python API already out there.

PR #217, opened 2022-01-13
Remove cogify

gadomski:
**Related Issue(s):** #216

**Description:** As @geospatial-jeff mentioned in https://github.com/stac-utils/stactools/issues/216#issuecomment-1012415230, our `cogify` function is a poor re-implementation of [rio-cogeo](https://github.com/cogeotiff/rio-cogeo). Since we aren't adding any value in **stactools**, this PR just removes our `cogify` function. Downstream libraries that were using `cogify` should switch to using **rio-cogeo**.

This is obviously a breaking change, so should be part of a minor release. 

**PR checklist:**


vincentsarago:
you could also use pure rasterio

```python
import rasterio
from rasterio.shutil import copy
    with rasterio.open("non_cog.tif") as src_dst:
        profile = src_dst.meta.copy()
        
        # update profile (block size, compression,...)
        profile["driver"] = "COG"
        profile["blocksize"] = 512. # example
        profile("blockxsize", None)
        profile("blockysize", None)
        
        copy(src_dst, "cog.tif", **profile)
```
ref: https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/cogeo.py#L352-L379

using rio-cogeo would still be valuable if you support rasterio version build with GDAL <3.1

PR #222, opened 2022-01-19
Remove GDAL's Python bindings

### LINKED ISSUE #216: Remove GDAL requirement

gadomski:
**Related Issue(s):** #216, #217

**Description:** To simplify our dependency tree, this PR removes direct use of the GDAL Python bindings (used in the `addraster` command). This represents a breaking change because it changes the function signature of `cogify`.

Additional fix:
- Refactors `cogify` to use **rasterio** instead of a system call to `gdal_translate`. This is _instead of_ #217 (and #217 should be closed if this PR is merged).

cc @sgillies
Closes #216 
Closes #217

**PR checklist:**


tests/core/test_utils.py
```diff
@@ -26,15 +26,9 @@ def test_default(self):
...
             self.assertTrue(os.path.exists(outfile))
             with rasterio.open(outfile) as dataset:
                 self.assertEqual(dataset.compression,
                                  rasterio.enums.Compression.lzw)
-
-    def test_extra_args(self):
-        with self.cogify(
-                extra_args=["-mo", "TIFFTAG_ARTIST=prince"]) as outfile:
-            with rasterio.open(outfile) as dataset:
-                self.assertEqual(dataset.tags()["TIFFTAG_ARTIST"], "prince")
```
sgillies, line 40: Formerly known as "prince" :purple_heart:

sgillies:
@gadomski I think the cogify function could be removed and left to rio-cogeo or gdal_translate, but this PR is fine. Rasterio's copy function calls [GDALCreateCopy](https://gdal.org/api/gdaldriver_cpp.html#classGDALDriver_1a00516dd36aa547771d316608237d9b52) and thus should be a pretty complete COG maker.

src/stactools/core/addraster.py
```diff
@@ -25,27 +27,29 @@ def add_raster_to_item(item: Item) -> Item:
...
+    bands = []
+    with rasterio.open(href) as dataset:
+        for (i, index) in enumerate(dataset.indexes):
+            data = dataset.read(index, masked=True)
+            band = RasterBand.create()
+            band.nodata = dataset.nodatavals[i]
+            band.spatial_resolution = dataset.transform[0]
+            band.data_type = DataType(dataset.dtypes[i])
+            minimum = float(numpy.min(data))  # type: ignore
+            maximum = float(numpy.max(data))  # type: ignore
```
duckontheweb, line 47:
```suggestion
            minimum = float(numpy.min(data))
            maximum = float(numpy.max(data))
```

I get `error: Unused "type: ignore" comment` when I run the lint script locally for these lines.

It looks like `scripts/lint` doesn't get run as part of the CI, is that intentional?

src/stactools/core/addraster.py
```diff
@@ -25,27 +27,29 @@ def add_raster_to_item(item: Item) -> Item:
...
+            band.statistics = Statistics.create(minimum=minimum,
+                                                maximum=maximum)
+            hist_data, _ = numpy.histogram(  # type: ignore
```
duckontheweb, line 50:
```suggestion
            hist_data, _ = numpy.histogram(
```

Same as above.

tests/cli/commands/test_addraster.py
```diff
@@ -39,4 +39,6 @@ def test_add_raster_to_item(self):
             asset = item.get_assets().get("analytic")
             assert asset is not None
             expected = expected_json("rasterbands.json")
-            assert expected == asset.to_dict().get("raster:bands")
+            self.maxDiff = None
+            for a, b in zip(expected, asset.to_dict().get("raster:bands")):
+                self.assertDictEqual(a, b)
```
duckontheweb, line 44: Not something that needs to be changed in this PR, but this might be a good use-case for [parametrized tests](https://docs.pytest.org/en/6.2.x/parametrize.html) now that we're using `pytest`.

duckontheweb (approved):
LGTM, I just had some comments around typing and possible parametrizing some tests. Nothing that should hold back this PR, though.

src/stactools/core/addraster.py
gadomski: We hit `scripts/cibuild` [here](https://github.com/stac-utils/stactools/blob/08f61bdf5f51db0dec4dbbeb7472a7b15091d722/.github/workflows/continuous-integration.yml#L58), which ends up calling `scripts/lint` [here](https://github.com/stac-utils/stactools/blob/08f61bdf5f51db0dec4dbbeb7472a7b15091d722/scripts/test#L20). I needed those `type: ignore` slugs to get things passing CI with our environment. Do you have more **numpy** typing available in your environment? If so can you point me at it? :-D

tests/cli/commands/test_addraster.py
gadomski: Cool, yeah I hadn't looked into that much before. Useful!

src/stactools/core/addraster.py
duckontheweb: > We hit `scripts/cibuild` [here](https://github.com/stac-utils/stactools/blob/08f61bdf5f51db0dec4dbbeb7472a7b15091d722/.github/workflows/continuous-integration.yml#L58), which ends up calling `scripts/lint` [here](https://github.com/stac-utils/stactools/blob/08f61bdf5f51db0dec4dbbeb7472a7b15091d722/scripts/test#L20). I needed those `type: ignore` slugs to get things passing CI with our environment. Do you have more **numpy** typing available in your environment? If so can you point me at it? :-D

Oops, sorry. I missed the call in `scripts/test` 🤦🏻 .

Looks like it's a difference in the `numpy` version. I was running `numpy==1.22.1` and didn't need those ignore statements. Going back to `numpy==1.21.1` (which is what is installed in the `conda` environment in the CI container) gives me errors without those ignore comments.

Issue #223, opened 2022-01-26
Remove Python 3.7 support


gadomski:
PySTAC will remove Python 3.7 support in stac-utils/pystac#730, and once that is released, we should follow suit. That change will require PySTAC to bump to v2.0, so we should similarly do a version bump (either to v0.3 or v1.0).

philvarner:
I'm in favor of doing this in stactools 0.4.0 regardless of what pystac does. There's likely very little use of 3.7 in the wild, since there aren't many legacy stac projects.

PR #228, opened 2022-02-08
Relax dependencies

### LINKED ISSUE #227: Relax upper bounds on dependencies
Currently, all the stactools dependencies are set as `~=` dependencies: https://github.com/stac-utils/stactools/blob/4a19089454fbe2a6148bf6e062a7f9393c636ca9/setup.cfg#L33-L42. Would folks be OK with removing those upper bounds?

For some packages that follow calver, like `fsspec`, that kind of constraint isn't meaningful. There's no reason to pin to a specific month's release.

It's also not meaningful for packages that use `x.y.z` version numbers, but don't actually promise to follow semver. I haven't checked if all those libraries declare that they do.

I also don't think it's appropriate for a "base" library like stactools to setup upper bounds. What if a downstream package requires a higher version of a package than what's allowed here?

And finally, IMO it's almost never appropriate for a library to declare an upper bound. Who's to say that stactools isn't compatible with the next release of that package? https://iscinumpy.dev/post/bound-version-constraints/ goes into this in much more detail, but the short version is that it's best for libraries to be as flexible as possible and let specific applications to pin their dependencies as they desire.

gadomski:
**Related Issue(s):**
- Closes #227 
- Relates to #223 in that we can't use `numpy>=1.22.0` until we drop Python 3.7, and until we do we need some `# type: ignore` comments on a couple of numpy calls that need to be ignored for newer numpy versions.

**Description:**
- Removes the upper bounds on our dependencies.
- Lowers the lower bounds on our dependencies, where possible. This process was pretty unscientific, mostly just manually installing versions and running the test suite.
- Adds `minimum-versions` and `pre-release-versions` to our CI.
- Changes the names of the python build actions
- Breaks `python-matrix` into `standard` and `extra-requires`
- Refactors the Python setup to repository-specific composite action to reduce duplication in the workflow file

Note that we had to unset `PROJ_LIB` to get the pre-release to pass, as described here: https://github.com/rasterio/rasterio/blob/master/docs/faq.rst#why-cant-rasterio-find-projdb-rasterio-from-pypi-versions--120. Not quite sure the correct approach to future-proof ourselves here, so going to just leave this as the solution in lieu of a better suggestion.

**PR checklist:**

- [x] Bring "required" actions into sync with the new actions/names (will be done after approval)

### DIFF
--- a/scripts/check_minimum_requirements
+++ b/scripts/check_minimum_requirements
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+# type: ignore
+
+import os.path
+import sys
+from importlib_metadata import requires
+from packaging.requirements import Requirement
+
+package_requirements = [
+    Requirement(requirement) for requirement in requires("stactools")
+]
+requirements_min = os.path.join(os.path.dirname(os.path.dirname(__file__)),
+                                "requirements-min.txt")
+with open(requirements_min) as file:
+    min_requirements = [Requirement(line) for line in file]
+min_requirements = dict(
+    (requirement.name, requirement) for requirement in min_requirements)
+
+incorrect_requirements = list()
+for package_requirement in package_requirements:
+    if package_requirement.marker is not None:
+        continue
+    min_requirement = min_requirements[package_requirement.name]
+    for (package_specifier,
+         min_specifier) in zip(package_requirement.specifier,
+                               min_requirement.specifier):
+        if package_specifier.operator == ">=" and package_specifier.version != min_specifier.version:
+            incorrect_requirements.append(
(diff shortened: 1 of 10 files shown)


scripts/check_minimum_requirements
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
```
TomAugspurger, line 1: What's the motivation for this script? As long as users (including devs) are `pip install`ing stactools, they shouldn't end up with too-old versions of any of the deps.

TomAugspurger (approved):
LGTM, just one minor question.

PR #258, opened 2022-03-08
skip self links check

jonhealy1:

**Related Issue(s):**
https://github.com/radiantearth/stac-spec/pull/1173

**Description:**
I am removing the warning for no self links.

**PR checklist:**


gadomski:
This feels like something that could be useful if pushed all the way down to `stac-check`. Two options I see right off the bat:
- Add a single `check_self_links: bool` parameter to the checker to enable/disable self link checks
- Add an `ignored_checks` `List` that could be used to disable a variety of checks on the checker

What do you think?

jonhealy1:
> This feels like something that could be useful if pushed all the way down to `stac-check`. Two options I see right off the bat:
> 
> * Add a single `check_self_links: bool` parameter to the checker to enable/disable self link checks
> * Add an `ignored_checks` `List` that could be used to disable a variety of checks on the checker
> 
> What do you think?

Both options sound good. Maybe an ignored checks list would best but checks having a boolean field is good too.

jonhealy1:
I think maybe we have to update how the config file is brought in from stac-check - I might just be confused

jonhealy1:
It seems to throw an error on the else statement and then never checks for the config file
```def parse_config(config_file):
        default_config_file = os.getenv("STAC_CHECK_CONFIG")
        if default_config_file:
            with open(default_config_file) as f:
                default_config = yaml.load(f, Loader=yaml.FullLoader)
        else:
            with pkg_resources.resource_stream(__name__, "stac-check.config.yml") as f:
                default_config = yaml.load(f, Loader=yaml.FullLoader)
        if config_file:
            with open(config_file) as f:
                config = yaml.load(f, Loader=yaml.FullLoader)
            default_config.update(config)
            
        return default_config

jonhealy1:
Maybe I wasn't setting the .env properly?

jonhealy1:
It is working now but the file is in an awkward place I think

gadomski:
I've pushed a commit to your branch that works around the issue in a different way -- we fall back on the default `stac-check` configuration, and simply disable the config value for the self links check. We also add a command line flag that allows the user to pass in their own configuration file, in case they want to re-enable it (or tweak the configuration any other way). I've requested your review to get your feeback on the idea.

jonhealy1:
@gadomski I don't see it - maybe it didn't push properly?

gadomski:
https://github.com/stac-utils/stactools/pull/258/commits/7b248fc189d017a61186df80574cb0fdc07775ff ?

jonhealy1:
Haha sorry I thought you meant push to my branch on stac-check

jonhealy1:
Looks great I think

gadomski:
Great thanks! I cleaned up the git history, will merge when CI passes. Thanks for the update!!

gadomski (approved):

PR #259, opened 2022-03-15
Add antimeridian helpers to utils

gadomski:
**Related Issue(s):** n/a

**Description:**
Includes some new helper functions:
- `normalize`: returns a new Polygon that has all of its longitudes in the same sign, even if this means going beyond -180 or 180.
- `split`: returns a new MultiPolygon that contains polygons on either side of the antimeridian, per the GeoJSON recommendations.
- `fix_item`: applies these change to an Item's geometry, inplace

**PR checklist:**

### DIFF
--- a/src/stactools/core/utils/antimeridian.py
+++ b/src/stactools/core/utils/antimeridian.py
@@ -33,10 +33,11 @@ def fix_item(item: Item, strategy: Strategy) -> None:
             f"Can only fix antimeridian issues for Polygons, geometry={geometry}"
         )
     if strategy == Strategy.NORMALIZE:
-        geometry = normalize(geometry)
-        bbox = geometry.bounds
-        item.geometry = shapely.geometry.mapping(geometry)
-        item.bbox = bbox
+        normalized_geometry = normalize(geometry)
+        if normalized_geometry:
+            bbox = normalized_geometry.bounds
+            item.geometry = shapely.geometry.mapping(normalized_geometry)
+            item.bbox = bbox
     elif strategy == Strategy.SPLIT:
         split_geometry = split(geometry)
         if split_geometry:
@@ -70,6 +71,8 @@ def split(polygon: Polygon) -> Optional[MultiPolygon]:
             no split occurred.
     """
     normalized = normalize(polygon)
+    if normalized is None:
+        return None
     if normalized.bounds[0] < -180:
         longitude = -180
     elif normalized.bounds[2] > 180:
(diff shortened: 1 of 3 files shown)


pjhartzell (approved):
Looks good. There is an edge case where it does not work as expected: polygons enclosing the north or south pole. Perhaps add that caveat in the docstring and put that down as a TODO.

gadomski:
Converting to draft until I implement the functionality described in https://github.com/stactools-packages/landsat/pull/13#discussion_r830182634.

gadomski:
@pjhartzell I've added `fix_item` if you want to take a look.

lossyrob:
I was expecting the behavior to be that if a geometry was more "westernly" than "easternly", then the normalization would use negative numbers instead of positive. E.g.

```python
def test_item_fix_antimeridian_normalize_west(self) -> None:
        canonical = Polygon(
            ((170, 40), (170, 50), (-140, 50), (-140, 40), (170, 40)))
        item = Item("an-id",
                    geometry=shapely.geometry.mapping(canonical),
                    bbox=canonical.bounds,
                    datetime=datetime.datetime.now(),
                    properties={})
        antimeridian.fix_item(item, antimeridian.Strategy.NORMALIZE)
        expected = shapely.geometry.box(170, 40, 190, 50)
        assert shapely.geometry.shape(item.geometry).equals(expected)
        assert item.bbox
        assert list(item.bbox) == [-190.0, 40.0, -140.0, 50.0]
```

This would allow e.g. STAC queries that are asking for western areas to pick up the appropriate geometries. As it stands it looks like anything that crosses the antimeridian from the west will get labeled as a very east polygon.

gadomski:
@lossyrob done in d8b9d410303f70b9c62148b0042eac92e25bf8bb.

src/stactools/core/utils/antimeridian.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Returns:
+        shapely.geometry.Polygon: The normalized polygon.
+    """
+    coords = list(polygon.exterior.coords)
+    for index, (start, end) in enumerate(zip(coords, coords[1:])):
+        delta = end[0] - start[0]
+        if abs(delta) > 180:
+            coords[index + 1] = (end[0] - math.copysign(360, delta), end[1])
+    polygon = Polygon(coords)
```
lossyrob, line 124: Maybe worth flagging if there are no changes, and returning `polygon` directly if not? Saves a couple of CPU cycles for creating a new Polygon from the coords, checking centroid, doing translate.

tests/core/test_utils.py
```diff
@@ -32,3 +38,126 @@ def test_profile(self):
             with rasterio.open(outfile) as dataset:
                 self.assertEqual(dataset.compression,
                                  rasterio.enums.Compression.lzw)
+
+    def test_antimeridian_split(self) -> None:
```
lossyrob, line 42: These tests currently live in "CogifyTest", which is off...can just make another TestCase in this same file specific to these utility methods.

lossyrob (approved):
Couple of small suggestions, but besides the known TODO this LGTM

tests/core/test_utils.py
gadomski: Since we're using pytest now I just moved them to the top level: 64986e779309817615519b358e53bc4a3bf6baf7.

src/stactools/core/utils/antimeridian.py
gadomski: Done 238798c, then with a refactor to allow quick short-circuiting of `split` too 3d9999a.

Issue #13, opened 2020-11-14
Show progressive console output during stac info 


lossyrob:
The `stac info` CLI command may take a bit to crawl the catalog. The goal of this issue is to show intermediate console output that will show progress to the user.

PR #262, opened 2022-03-18
Progressive console output for `stac info`

rushgeo:
**Related Issue(s):**
#13 

**Description:**
Added `-p` & `--progress` options for `stac info`. The output is overwritten on every step through the catalog. Extensions are only printed at the end. 

I didn't see any difference in runtime when testing against the largest catalogs in this repo's `tests/data-files` dir. 

**PR checklist:**

### DIFF
--- a/src/stactools/cli/commands/info.py
+++ b/src/stactools/cli/commands/info.py
@@ -22,7 +22,7 @@ def print_info(
         nonlocal max_line_length
         line = (
             f"\rCatalogs={cat_count}, Collections={col_count}, "
-            "Items={item_count}: Reading {root}"
+            f"Items={item_count}: Reading {root}"
         )
         if len(line) > max_line_length:
             max_line_length = len(line)
--- a/src/stactools/cli/commands/info.py
+++ b/src/stactools/cli/commands/info.py
@@ -1,15 +1,33 @@
 import click
 import pystac
+from pystac import Catalog, STACObject
 
 
-def print_info(catalog_path: str, skip_items: bool = False) -> None:
+def print_info(
+    catalog_path: str, skip_items: bool = False, print_progress: bool = False
+) -> None:
     cat_count, col_count, item_count = 0, 0, 0
     cat_ext, col_ext, item_ext = set([]), set([]), set([])
 
     cat = pystac.read_file(catalog_path)
-    if not isinstance(cat, pystac.Catalog):
+    if not isinstance(cat, Catalog):
         print("{} is not a catalog".format(catalog_path))
         return
+
+    cat_id_info = "Catalog ID: {}".format(cat.id)
+
+    max_line_length = 0
+
+    def progress_print(root: STACObject) -> None:
+        nonlocal max_line_length
+        line = (
+            f"\rCatalogs={cat_count}, Collections={col_count}, "
+            "Items={item_count}: Reading {root}"
+        )
+        if len(line) > max_line_length:
+            max_line_length = len(line)
+        print("\r" + " " * max_line_length, end="")
+        print(line, end="")
 
     for root, _, items in cat.walk():
         if root.STAC_OBJECT_TYPE == pystac.STACObjectType.COLLECTION:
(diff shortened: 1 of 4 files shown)


src/stactools/cli/commands/info.py
```diff
@@ -1,8 +1,14 @@
...
+import sys
+from colorama import init  # type: ignore
 
+init()
 
-def print_info(catalog_path: str, skip_items: bool = False) -> None:
+
+def print_info(catalog_path: str,
+               skip_items: bool = False,
+               prog_print: bool = False) -> None:
```
gadomski, line 11:
```suggestion
               print_progress: bool = False) -> None:
```

requirements-min.txt
```diff
@@ -9,3 +9,4 @@ pystac[validation] == 1.2
 rasterio == 1.2.9
 requests == 2.20
 stac-check == 1.1.3
+colorama == 0.4.4
```
gadomski, line 12: So this `requirements-min.txt` file is only used for CI (which makes me wonder if we should move it into the `.github` folder 🤔). To specify as a requirements, this needs to be added to `setup.cfg`.

Also, since this is a Windows-only requirement, we can guard the install w/ something like this: `colorma >= 0.4; sys_platform = 'win32'` (or something, see https://stackoverflow.com/a/35614580/732529).

gadomski (changes requested):

gadomski:
@rushgeo fysa I rebase onto **main** so you'll need a hard reset to take up the new changes.

pjhartzell (commented):
1. No tests. Maybe OK for something like this.
2. As things progress, the overwrite does not clear the line, so you end up with detritus at the end of the line. See below.

**First "Reading" output is OK:**
  ```
~/dev/stactools proginfo*
  stactools ❯ stac info tests/data-files/catalogs/planet-example-1.0.0-beta.2/collection.json --progress
  Catalog ID: planet-disaster-data
  --------------------------------
     CATALOGS: 0
  COLLECTIONS: 1
        ITEMS: 0
  Reading <Collection id=planet-disaster-data>...
  ```
**Second "Reading" output has leftovers from the first output at the end (bottom line):**
```
~/dev/stactools proginfo*
stactools ❯ stac info tests/data-files/catalogs/planet-example-1.0.0-beta.2/collection.json --progress
Catalog ID: planet-disaster-data
--------------------------------
   CATALOGS: 1
COLLECTIONS: 1
      ITEMS: 0
Reading <Catalog id=hurricane-harvey>...ata>...
```

3. When it is done, the "Reading" status for the last catalog is still shown. Feels kinda clunky. Maybe clear that line when done, or write a message indicating it's done.

**Finishes with message that it is reading the last catalog rather than clearing the status:**
```
~/dev/stactools proginfo*
stactools ❯ stac info tests/data-files/catalogs/planet-example-1.0.0-beta.2/collection.json --progress
Catalog ID: planet-disaster-data
--------------------------------
   CATALOGS: 2 
COLLECTIONS: 1 
      ITEMS: 5 (extensions: https://stac-extensions.github.io/eo/v1.0.0/schema.json,https://stac-extensions.github.io/projection/v1.0.0/schema.json,https://stac-extensions.github.io/view/v1.0.0/schema.json)
Reading <Catalog id=hurricane-harvey-0831>.....
~/dev/stactools proginfo*
stactools ❯ 
```

gadomski:
> No tests. Maybe OK for something like this.

Nah, we should do tests. Adding.

However, I'm not seeing the same issues as you w.r.t the progressive output. I think the "raw" way that we're doing stdoutput manipulation might be a little fragie -- shells and output are hard. I'll see if there's something in `click` or elsewhere that we can use to take care of these fiddly bits for us.

gadomski:
@pjhartzell I've added some tests, and now the progress only works on one line, which _should_ be more robust across terminals. I had to fix a couple of things in one of the tests files as well.

src/stactools/cli/commands/info.py
```diff
@@ -1,16 +1,34 @@
...
 
+    cat_id_info = "Catalog ID: {}".format(cat.id)
+
+    max_line_length = 0
+
+    def progress_print(root: STACObject) -> None:
+        nonlocal max_line_length
+        line = (
+            f"\rCatalogs={cat_count}, Collections={col_count}, "
+            "Items={item_count}: Reading {root}"
```
pjhartzell: Need an `f` in front of `"Items={item_count}: Reading {root}"` in order to see the variable values:

`f"Items={item_count}: Reading {root}"`

Looks good otherwise!

pjhartzell (approved):
I fixed a minor typo that affected the output, looks good.

src/stactools/cli/commands/info.py
gadomski: good catch thanks!

gadomski (approved):

PR #268, opened 2022-03-30
Large-scale refactor

gadomski:
**Related Issue(s):** None

**Description:** This is a large-scale _DRAFT_ pull request to experiment with ideas for refactoring stactools for the next major release (v0.4.0). Some goals:
- Provide a clean interface for new dataset packages to easily create STAC objects with the minimal amount of boilerplate while following best practices
- Consolidate/reorganize the various utility methods scattered around `stactools.core`
- Provide a clean interface for command-line functions to reduce the amount of boilerplate required to implement those
- Other goals TBD

**PR checklist:**

- [ ] Code is formatted (run `scripts/format`).
- [ ] Code lints properly (run `scripts/lint`).
- [ ] Tests pass (run `scripts/test`).


docs/builder.rst
```diff
@@ -0,0 +1,123 @@
...
+
+The base class, :py:class:`stactools.core.builder.Builder`, is pretty useless on
+its own. It takes a dictionary of assets as its initializer, and wraps those
+assets in an Item with a provided id and an auto-generated datetime::
+
+    from stactools.core.builder import Builder
+    from pystac import Asset
+
+    builder = Builder()
+    builder.add_asset("an-asset", Asset(href="a-file.dat"))
```
TomAugspurger, line 21: Style question: what's the motivation for `builder = Builder(); builder.add_asset` pattern rather than defining what `Builder.create_item` will do in the class definitation?

TomAugspurger (commented):
I mentioned this on a call earlier, but this is a pattern built on fsspec that I've found helpful.

https://github.com/TomAugspurger/era5/blob/main/src/stactools/era5/stac.py#L150-L155

The `create_item` includes the `protocol` and `storage_options` keywords.

```python
def create_item(
    path: str,
    kind: str,
    protocol: str,
    storage_options: dict[str, Any] | None = None,
) -> pystac.Item:
```

And then the usage internally is

```python
    storage_options = storage_options or {}
    fs = fsspec.filesystem(protocol, **storage_options)
```

And all paths are relative to the dataset root. All path-like operations within `create_item` would use fsspec. *In theory*, this lets you use the same code for STAC items that might live on different filesystems (S3, Azure Blob Storage, local, ...). Hope that's helpful :)

docs/builder.rst
gadomski: Partially, it's my procedural background, but I do find that leaving the asset addition until later is useful for conditionals, see here: https://github.com/stactools-packages/modis/blob/524bfd26bb6d571336e50a226e853b97287529e3/src/stactools/modis/commands.py#L74-85.

PR #274, opened 2022-04-05
Format with black

gadomski:
**Related Issue(s):**
- https://github.com/stac-utils/pystac/issues/316

**Description:**
We generally defer to PySTAC when it comes to tooling/formatting/linting decisions, and PySTAC uses black. The changes in `src/` and `test/` are _only_ from running `scripts/format`, there are no functional changes.

I'm queuing up another PR to introduce pre-commit and isort, but I'm doing this PR separately since there's so much formatting noise in the diff.

**PR checklist:**

- [x] Code is formatted (run `scripts/format`).
- [x] Code lints properly (run `scripts/lint`).
- [x] Tests pass (run `scripts/test`).
- [x] Documentation has been updated to reflect changes, if applicable.
- [x] Changes are added to the [CHANGELOG](../CHANGELOG.md).

PR #275, opened 2022-04-05
Add pre-commit and isort

gadomski:
**Related Issue(s):**
- Depends on #274 (marking this as DRAFT until that is merged)
- https://github.com/stac-utils/pystac/pull/401
- https://github.com/stac-utils/pystac/issues/533

**Description:**
Adds pre-commit -- PySTAC uses it, so we should too. We don't follow PySTAC's configuration exactly, because they use local python for the repos in the pre-commit configuration, whereas we use the repos themselves. PySTAC doesn't use isort yet but it might, and I think its handy because I dislike having to think about import ordering.

There was one typing fix:
- Had to cast a return in `stactools.core.io.xml` to get mypy happy

Otherwise no functional changes, just a lot of re-ordering noise.

I updated the README to reference pre-commit, and made some other README updates while I was in there.

**PR checklist:**


philvarner:
lol, I hadn't seen this when I made my comment on the other ticket.

docker/Dockerfile
```diff
@@ -31,6 +31,7 @@ RUN conda install -c conda-forge pandoc && conda clean -af
 COPY requirements-dev.txt ./
 RUN pip install -r requirements-dev.txt
 COPY . ./
+RUN git init
```
philvarner, line 34: why is git init run?

philvarner (approved):

Issue #147, opened 2021-06-22
Add an asset to an item


cholmes:
One of the more common things I've been doing is merging collections in order to combine their assets. It would be quite useful to just have a command that operates at the item level, that lets you add in an asset. Would just take an item and a file as input, with optional parameters to specify the role, type, etc.

gadomski:
Closed by #300.

PR #300, opened 2022-06-07
feat: command to add an asset to an item

sunu:
**Related Issue(s):**
https://github.com/stac-utils/stactools/issues/147

**Description:**
Adds a command called `addasset` to add an asset to an item. It takes an item's path, an asset's key, path and optionally roles, type etc as argument. The command doesn't handle `extra_fields` yet. 

**PR checklist:**

### DIFF
--- a/src/stactools/cli/commands/add_asset.py
+++ b/src/stactools/cli/commands/add_asset.py
@@ -2,6 +2,7 @@ from typing import List, Optional
 
 import click
 import pystac
+import pystac.utils
 
 from stactools.core import add_asset_to_item
 
@@ -78,9 +79,9 @@ def create_add_asset_command(cli: click.Group) -> click.Command:
         ignore_conflicts: bool = False,
     ) -> None:
         _add_asset(
-            item_path,
+            pystac.utils.make_absolute_href(item_path),
             asset_key,
-            asset_path,
+            pystac.utils.make_absolute_href(asset_path),
             title,
             description,
             media_type,
--- a/src/stactools/cli/commands/add_asset.py
+++ b/src/stactools/cli/commands/add_asset.py
@@ -1,0 +1,92 @@
+from typing import List, Optional
+
+import click
+import pystac
+
+from stactools.core import add_asset_to_item
+
+
+def _add_asset(
+    item_path: str,
+    asset_key: str,
+    asset_path: str,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+    media_type: Optional[str] = None,
+    roles: Optional[List[str]] = None,
+    move_assets: bool = False,
+    ignore_conflicts: bool = False,
+) -> None:
+    item = pystac.read_file(item_path)
+    if not isinstance(item, pystac.Item):
+        raise click.BadArgumentUsage(f"{item_path} is not a STAC Item")
+    asset = pystac.Asset(asset_path, title, description, media_type, roles)
+    item = add_asset_to_item(
+        item,
+        asset_key,
+        asset,
+        move_assets=move_assets,
+        ignore_conflicts=ignore_conflicts,
+    )
+    item.save_object()
+
+
+def create_add_asset_command(cli: click.Group) -> click.Command:
+    @cli.command("add-asset", short_help="Add an asset to an item.")
+    @click.argument("item_path")
+    @click.argument("asset_key")
+    @click.argument("asset_path")
+    @click.option("--title", help="Optional title of the asset")
+    @click.option(
+        "--description",
+        help=(
+            "Optional description of the asset providing additional details, "
+            "such as how it was processed or created."
+        ),
+    )
+    @click.option("--media-type", help="Optional media type of the asset")
+    @click.option(
+        "-r",
+        "--role",
+        "roles",
+        help="Optional, semantic roles of the asset",
+        multiple=True,
+    )
+    @click.option(
+        "--move-assets",
+        is_flag=True,
+        help="Move asset to the target Item's location.",
+    )
+    @click.option(
+        "--ignore-conflicts",
+        is_flag=True,
+        help=(
+            "If there already exists an asset with the given key or at the "
+            "target path (when `--move-assets` flag is set), do not raise an "
+            "error, leave the original asset from the target item in place."
+        ),
+    )
+    def add_asset_command(
+        item_path: str,
+        asset_key: str,
+        asset_path: str,
+        title: Optional[str] = None,
+        description: Optional[str] = None,
+        media_type: Optional[str] = None,
+        roles: Optional[List[str]] = None,
+        move_assets: bool = False,
+        ignore_conflicts: bool = False,
+    ) -> None:
+        _add_asset(
+            item_path,
+            asset_key,
+            asset_path,
+            title,
+            description,
+            media_type,
+            roles,
+            move_assets=move_assets,
+            ignore_conflicts=ignore_conflicts,
+        )
+
+    return add_asset_command
--- a/src/stactools/core/add_asset.py
+++ b/src/stactools/core/add_asset.py
@@ -1,0 +1,66 @@
+import logging
+
+from pystac import Asset, Item
+from pystac.utils import is_absolute_href, make_relative_href
+
+from stactools.core.copy import move_asset_file_to_item
+
+logger = logging.getLogger(__name__)
+
+
+def add_asset_to_item(
+    item: Item,
+    key: str,
+    asset: Asset,
+    move_assets: bool = False,
+    ignore_conflicts: bool = False,
+) -> Item:
+    """Adds an asset to an item.
+
+    Args:
+        item (Item): The PySTAC Item to which the asset will be added.
+        key (str): The unique key of the asset.
+        asset (Asset): The PySTAC Asset to add.
+        move_assets (bool): If True, move the asset file alongside the target item.
+        ignore_conflicts (bool): If True, asset with the same key will not be added,
+            and asset file that would overwrite an existing file will not be moved.
+            If False, either of these situations will throw an error.
+
+    Returns:
+        Item: Returns an updated Item with the added Asset.
+            This operation mutates the Item.
+    """
+    item_href = item.get_self_href()
+    asset_href = asset.get_absolute_href()
+    if key in item.assets:
+        if not ignore_conflicts:
+            raise Exception(
+                f"Target item {item} already has an asset with key {key}, "
+                "cannot add asset in from {asset_href}"
+            )
+    else:
+        if not asset_href:
+            raise ValueError(
+                f"Asset {asset} must have an href to be added. The href "
+                "value should be an absolute path or URL."
+            )
+        if not item_href and move_assets:
+            raise ValueError(
+                f"Target Item {item} must have an href to move an asset to the item"
+            )
+        if not item_href and not is_absolute_href(asset.href):
+            raise ValueError(
+                f"Target Item {item} must have an href to add "
+                "an asset with a relative href"
+            )
+        if move_assets:
+            new_asset_href = move_asset_file_to_item(
+                item, asset_href, ignore_conflicts=ignore_conflicts
+            )
+        else:
+            if not is_absolute_href(asset.href) and item_href is not None:
+                asset_href = make_relative_href(asset_href, item_href)
+            new_asset_href = asset_href
+        asset.href = new_asset_href
+        item.add_asset(key, asset)
+    return item
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Conda package, via [conda-forge](https://anaconda.org/conda-forge/stactools) ([#324](https://github.com/stac-utils/stactools/pull/324))
 - Context manager to ignore rasterio's NotGeoreferencedWarning ([#331](https://github.com/stac-utils/stactools/pull/331))
 - Added `raster_footprint` module to assist in populating the geometry of an Item from data coverage of its data assets ([#307](https://github.com/stac-utils/stactools/pull/307))
+- `stac add-asset` command to add an asset to an item ([#300](https://github.com/stac-utils/stactools/pull/300))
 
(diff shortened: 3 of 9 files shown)


sunu:
The failing tests on Python 3.10 seem to be unrelated to this PR. I've created a separate PR (https://github.com/stac-utils/stactools/pull/301) to handle that.

tests/cli/commands/test_addasset.py
```diff
@@ -0,0 +1,113 @@
...
+    return temp_path
+
+
+class AddAssetTest(CliTestCase):
+    def create_subcommand_functions(self):
+        return [create_addasset_command]
+
+    def test_add_asset_to_item(self):
+        with TemporaryDirectory() as tmp_dir:
+            catalog = create_temp_catalog_copy(tmp_dir)
```
gadomski, line 38: I believe the `catalog` part of this test is not necessary, since we are just testing the behavior on the `Item`. The test should be reduced to only include the components necessary to test the expected behavior.

src/stactools/core/addasset.py
```diff
@@ -0,0 +1,57 @@
...
+            and asset file that would overwrite an existing file will not be moved.
+            If False, either of these situations will throw an error.
+
+    Returns:
+        Item: Returns an updated Item with the added Asset.
+            This operation mutates the Item.
+    """
+    item_href = item.get_self_href()
+    if item_href is None:
+        raise ValueError(f"Target Item {item.id} must have an HREF to add an asset")
```
gadomski, line 35: If `move_assets` is `False` and the asset href is absolute, I don't see any reason to error if the item doesn't have an href.

gadomski (changes requested):

src/stactools/cli/commands/addasset.py
```diff
@@ -0,0 +1,89 @@
...
+        "--move-assets",
+        is_flag=True,
+        help="Move asset to the target Item's location.",
+    )
+    @click.option(
+        "--ignore-conflicts",
+        is_flag=True,
+        help=(
+            "If there is a conflict with an existing asset, do not raise an "
+            "error, leave the original asset from the target item in place."
```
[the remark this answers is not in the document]
sunu: Yup, that's exactly what it means. I'll update description to be clearer.

src/stactools/core/addasset.py
```diff
@@ -0,0 +1,57 @@
...
+    if key in item.assets:
+        if not ignore_conflicts:
+            raise Exception(
+                "Target item {} already has asset with key {}, "
+                "cannot add asset in from {}".format(item, key, asset_href)
+            )
+    else:
+        asset_href = asset.get_absolute_href()
+        if asset_href is None:
+            raise ValueError(f"Asset {asset.title} must have an HREF to be added")
```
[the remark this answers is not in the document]
sunu: That's true. I think using `repr(asset)` would be better.

tests/cli/commands/test_addasset.py
```diff
@@ -0,0 +1,113 @@
...
+    return col
+
+
+def create_temp_copy(src_path, tmp_dir, target_name):
+    temp_path = os.path.join(tmp_dir, target_name)
+    shutil.copyfile(src_path, temp_path)
+    return temp_path
+
+
+class AddAssetTest(CliTestCase):
```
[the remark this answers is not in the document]
sunu: Yes, makes sense

src/stactools/cli/commands/addasset.py
```diff
@@ -0,0 +1,89 @@
...
+    asset_path: str,
+    title: Optional[str] = None,
+    description: Optional[str] = None,
+    media_type: Optional[str] = None,
+    roles: Optional[List[str]] = None,
+    move_assets: bool = False,
+    ignore_conflicts: bool = False,
+) -> None:
+    item = pystac.read_file(item_path)
+    if not isinstance(item, pystac.Item):
```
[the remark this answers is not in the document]
sunu: I briefly looked into what's required to support both an `Item` or a `Collection` as the target. From what I understand, we would need an equivalent of the `move_asset_file_to_item` function for `Collections`: https://github.com/stac-utils/stactools/blob/1077ccd58cafc4dd392abddefc490af2ee64c1ea/src/stactools/core/copy.py#L14

How do you feel about opening an issue for that and adding support for `Collections` in a separate PR?

sunu:
Hi @gadomski, I think I have addressed most of your comments and this should be ready for another look. Also, thank you for taking the time to providing so much feedback. I'm pretty new to STAC. So I really appreciate it 🙂

tests/core/test_addasset.py
```diff
@@ -0,0 +1,74 @@
...
+            assert asset is None
+            item_path = item.get_self_href()
+
+            asset_path = test_data.get_path("data-files/core/byte.tif")
+            asset = pystac.Asset(
+                asset_path, "test", "placeholder asset", roles=["thumbnail", "overview"]
+            )
+            item = add_asset_to_item(item, "test-asset", asset)
+
+            asset = item.get_assets().get("test-asset")
```
gadomski, line 32:
```suggestion
            asset = item.assets["test-asset")]
```

tests/core/test_addasset.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class AddAssetTest(TestCase):
+    def test_add_asset_to_item(self):
+        with TemporaryDirectory() as tmp_dir:
+            item_path = create_temp_copy(
+                test_data.get_path("data-files/core/simple-item.json"),
+                tmp_dir,
+                "item.json",
+            )
```
gadomski, line 19: Since we're not moving the asset, do we need to create a copy of the item in this unit test?

src/stactools/core/addasset.py
```diff
@@ -0,0 +1,62 @@
...
+        if asset_href is None:
+            raise ValueError(
+                f"Asset {repr(asset)} must have an absolute HREF to be added"
+            )
+        if item_href is None and (move_assets or not is_absolute_href(asset.href)):
+            # An absolute href for the item is required if we need to move the
+            # asset into the item or if the asset href is relative
+            raise ValueError(
+                f"Target Item {repr(item)} must have an HREF to add an asset"
+            )
```
gadomski, line 51: Syntax isn't perfect but you get the idea

```suggestion
        if item_href is None and move_assets:
            # An absolute href for the item is required if we need to move the
            # asset into the item or if the asset href is relative
            raise ValueError(
                f"Target Item {repr(item)} must have an HREF to move an asset to the item"
            )
        elif item_href is None and not is_absolute_href(asset.href):
            raise ValueError(f"Target Item {repr(item)} must have an href to add an asset with a relative href")
```

tests/core/test_addasset.py
```diff
@@ -0,0 +1,74 @@
...
+
+            asset = item.get_assets().get("test-asset")
+            assert isinstance(asset, pystac.Asset), asset
+            assert asset.href is not None, asset.to_dict()
+            assert os.path.isfile(asset.href), asset.to_dict()
+            assert asset.title == "test", asset.to_dict()
+            assert asset.description == "placeholder asset", asset.to_dict()
+            self.assertListEqual(asset.roles, ["thumbnail", "overview"])
+
+    def add_asset_move(self):
```
gadomski, line 40:
```suggestion
    def test_add_asset_move(self):
```

Otherwise, it isn't picked up as a test.

tests/core/test_addasset.py
```diff
@@ -0,0 +1,74 @@
...
+                asset = item.get_assets().get("test-asset")
+                assert isinstance(asset, pystac.Asset), asset
+                assert asset.href is not None, asset.to_dict()
+                assert os.path.isfile(asset.href), asset.to_dict()
+                assert asset.title == "test", asset.to_dict()
+                assert asset.description == "placeholder asset", asset.to_dict()
+                self.assertEqual(
+                    os.path.dirname(asset.get_absolute_href()),
+                    os.path.dirname(item.get_self_href()),
+                )
```
gadomski, line 74: This is a bit of a style thing, but a lot of these assertions are covered by `test_add_asset_to_item`, so don't need to be made twice. This could be simplified to something like:

```suggestion
        with TemporaryDirectory() as tmp_dir:
            item_path = create_temp_copy(
                test_data.get_path("data-files/core/simple-item.json"),
                tmp_dir,
                "item.json",
            )
            item = pystac.read_file(item_path)
            with TemporaryDirectory as tmp_dir2:
                asset_path = create_temp_copy(
                    test_data.get_path("data-files/core/byte.tif"), tmp_dir2, "test.tif"
                )
                asset = pystac.Asset(
                    asset_path,
                    "test",
                    "placeholder asset",
                    roles=["thumbnail", "overview"],
                )
                item = add_asset_to_item(
                    item, "test-asset", asset, move_assets=True, ignore_conflicts=True
                )

                asset = item.assets["test-asset"]
                assert isinstance(asset, pystac.Asset), asset
                assert asset.href is not None, asset.to_dict()
                assert os.path.isfile(asset.href), asset.to_dict()
                self.assertEqual(
                    os.path.dirname(asset.get_absolute_href()),
                    os.path.dirname(item.get_self_href()),
                )
```

tests/core/test_addasset.py
```diff
@@ -0,0 +1,74 @@
...
+from unittest import TestCase
+
+import pystac
+
+from stactools.core import add_asset_to_item
+from tests import test_data
+from tests.utils import create_temp_copy
+
+
+class AddAssetTest(TestCase):
```
gadomski, line 12: Thanks for adding some `core` unit tests! I think there are a couple more cases that would be useful to assert:

1. Asserting that an error is raised if `move=True` and the item doesn't have an href
2. Asserting that things work ok if the item doesn't have an href, `move=False`, and the asset href is absolute
3. Asserting an error if the item doesn't have an href, `move=False`, and the asset href is relative

There may be a couple others. When reading your tests, I was curious what happened in the above cases, and had to go read the code to find out.

gadomski (changes requested):

src/stactools/cli/commands/addasset.py
```diff
@@ -0,0 +1,92 @@
...
+        asset_key,
+        asset,
+        move_assets=move_assets,
+        ignore_conflicts=ignore_conflicts,
+    )
+    item.save_object()
+
+
+def create_addasset_command(cli: click.Group) -> click.Command:
+    @cli.command("addasset", short_help="Add an asset to an item.")
```
philvarner, line 35: We should establish a convention for multi-word commands. We currently only have two cases: `create-item` and `addraster`. I think we should standardize on kebab-case, and change addraster. @gadomski

src/stactools/core/addasset.py
```diff
@@ -0,0 +1,62 @@
...
+    Returns:
+        Item: Returns an updated Item with the added Asset.
+            This operation mutates the Item.
+    """
+    item_href = item.get_self_href()
+    asset_href = asset.get_absolute_href()
+    if key in item.assets:
+        if not ignore_conflicts:
+            raise Exception(
+                "Target item {} already has asset with key {}, "
```
philvarner, line 38: I would prefer f-string use here rather than format

src/stactools/core/addasset.py
```diff
@@ -0,0 +1,62 @@
...
+                "cannot add asset in from {}".format(item, key, asset_href)
+            )
+    else:
+        if asset_href is None:
```
philvarner, line 42:
```suggestion
        if not asset_href:
```

to also handle the case where asset_href is the empty string

src/stactools/cli/commands/addasset.py
gadomski: 👍🏽 to kebab case, ticket created: https://github.com/stac-utils/stactools/issues/322.

tests/core/test_addasset.py
sunu: oops, I totally overlooked this one.

tests/core/test_addasset.py
sunu: I have added tests for these cases and a few more.

sunu:
@gadomski @philvarner I have updated the code to address the review comments.

src/stactools/cli/commands/add_asset.py
```diff
@@ -0,0 +1,92 @@
...
+        description: Optional[str] = None,
+        media_type: Optional[str] = None,
+        roles: Optional[List[str]] = None,
+        move_assets: bool = False,
+        ignore_conflicts: bool = False,
+    ) -> None:
+        _add_asset(
+            item_path,
+            asset_key,
+            asset_path,
```
gadomski, line 83:
```suggestion
            pystac.utils.make_absolute_href(item_path),
            asset_key,
            pystac.utils.make_absolute_href(asset_path),
```

src/stactools/core/add_asset.py
```diff
@@ -0,0 +1,66 @@
...
+        ignore_conflicts (bool): If True, asset with the same key will not be added,
+            and asset file that would overwrite an existing file will not be moved.
+            If False, either of these situations will throw an error.
+
+    Returns:
+        Item: Returns an updated Item with the added Asset.
+            This operation mutates the Item.
+    """
+    item_href = item.get_self_href()
+    asset_href = asset.get_absolute_href()
```
gadomski, line 34:
```suggestion
    asset_href = asset.href
```

gadomski (changes requested):
I tried the following command and got an error:

```shell
$ stac add-asset tests/data-files/core/simple-item.json foobar tests/data-files/core/byte.tif
Traceback (most recent call last):
  <snip>
ValueError: Asset <Asset href=tests/data-files/core/byte.tif> must have an href to be added. The href value should be an absolute path or URL.
```

I think my proposed use-case (two relative, local hrefs) should be supported, so I've suggested changes that fix the error. Let me know what you think.

gadomski (approved):

pjhartzell (approved):

gadomski:
I've rebased this branch onto **main** and consolidated @sunu's contributions to one commit.

sunu:
Thank you for finishing the PR @gadomski! Sorry for dropping the ball in the middle.

gadomski:
No problem, thanks for the contribution!

PR #301, opened 2022-06-08
build: fix failing ci builds for python 3.10

sunu:
Set strict channel priority for conda.

**Description:**
The CI build for Python 3.10 is failing because of broken GDAL installation. Setting strict channel priority fixes the package compatibility problem (ref: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority)

Examples of failing builds:
- https://github.com/stac-utils/stactools/runs/6549941864?check_suite_focus=true#step:5:71
- https://github.com/stac-utils/stactools/runs/6778770957?check_suite_focus=true#step:5:78

**PR checklist:**

- [na] Documentation has been updated to reflect changes, if applicable.

### DIFF
--- a/.github/setup/action.yml
+++ b/.github/setup/action.yml
@@ -34,6 +34,9 @@ runs:
       with:
         auto-update-conda: true
         python-version: ${{ inputs.python-version }}
+    - name: Set strict channel priority
+      run: conda config --set channel_priority strict
+      shell: bash -l {0}
     - name: Update Conda's environemnt
       run: conda env update -f environment.yml -n test
       shell: bash -l {0}
--- a/.github/setup/action.yml
+++ b/.github/setup/action.yml
@@ -34,9 +34,6 @@ runs:
       with:
         auto-update-conda: true
         python-version: ${{ inputs.python-version }}
-    - name: Set strict channel priority
-      run: conda config --set channel_priority strict
-      shell: bash -l {0}
     - name: Update Conda's environemnt
       run: conda env update -f environment.yml -n test
       shell: bash -l {0}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Set strict channel priority for conda to reduce package incompatibility ([#301](https://github.com/stac-utils/stactools/pull/301))
+
 ## [v0.3.1]
 
 ### Added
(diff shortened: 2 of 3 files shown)


.github/setup/action.yml
```diff
@@ -34,6 +34,9 @@ runs:
       with:
         auto-update-conda: true
         python-version: ${{ inputs.python-version }}
+    - name: Set strict channel priority
+      run: conda config --set channel_priority strict
+      shell: bash -l {0}
```
gadomski, line 39: This works and I'm 👍🏽, but I was curious on your thoughts about using the `conda_forge::gdal` (e.g.) syntax in `enviornment.yml` instead, as described [here](https://stackoverflow.com/a/65983247/732529). This would (I think) also fix Python 3.10 for users installing on their local systems.

sunu: Hey @gadomski! I saw that thread while looking for possible fixes and in theory that should work on it's own without making any changes to `channel_priority`. But I couldn't get it to work. GDAL installation was still broken.

Here's what I have tried:
```
$ docker run -it -v $(pwd):/stactools continuumio/miniconda3 bash
$ conda create -n test python=3.10
$ conda activate test
$ conda env update -f /stactools/environment.yml  -n test
```
Running `gdalinfo` after these commands results in
```
(test) root@7d8865355e4d:/# gdalinfo
gdalinfo: /opt/conda/envs/test/bin/../lib/./libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /opt/conda/envs/test/bin/../lib/libgdal.so.30)
gdalinfo: /opt/conda/envs/test/bin/../lib/./libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /opt/conda/envs/test/bin/../lib/./libtiledb.so.2.8)
```
So the installation is broken.

This is what the `environment.yml` file looks like:
```
name: stactools
channels:
  - conda-forge
  - defaults
dependencies:
  - conda-forge::gdal>=3.3
  - conda-forge::geos>=3.3
  - conda-forge::rasterio>=1.2
```
fwiw, setting channel priority to strict and running `conda env update ...` again fixes the installation. But I don't exactly know why explicit channel namespace doesn't work on its own.

sunu: I think I have figured out what was causing the issue. `libstdcxx-ng` package was being installed from default channel as part of python 3.10 installation. Forcing that to be installed from the `conda-forge` channel fixes the issue.

I have updated the PR with that change.

gadomski (approved):

Issue #85, opened 2021-03-31
Add method for generating data footprints


matthewhanson:
I posted this https://github.com/stac-utils/stactools/pull/84#issuecomment-810697975 for Sentinel-1 and https://github.com/stac-utils/stactools/pull/83 for Landsat, but seems like it would be a good addition to stactools to be able to generate footprints when there are none provided with the original metadata.

*****
I spent some time looking into this for Sentinel-1 a while back and used rasterio to get the footprints.  I also tried using overviews to generate the footprints so it would be faster and wouldn't require reading as much of the file, but it really worked quite a bit better using the full resolution file.

The resulting geometry tends to be very large and way too big, so you have to simplify it.  It's a balancing act to simplify it enough to you still have the detail needed while keeping the geometry size small.

You can see in the gifs below the difference between the provided geometry (blue) and the one calculated with the above code (red)

This is a typical image and the provided boundary overestimates the area cross-track, and underestimates it along-track.

![sentinel1-boundaries-small](https://user-images.githubusercontent.com/1416900/113078983-31bafe80-91a2-11eb-91de-5516ff8890b7.gif)

This is a more unusual collect that is along the water where some of the water was masked out in the data, and we see the situation is even worse.  The generated footprint in red is far better.

![sentinel1-boundaries-odd-small](https://user-images.githubusercontent.com/1416900/113079122-79418a80-91a2-11eb-86aa-9a563a6a1159.gif)

There might be edge cases where this would fail, thus before going operational I think it should be used to generate a few hundred boundaries to be visually inspected.

If the image were already downloaded or locally available, it's not that bad to run this processing and certainly worth it for the better footprint.

The code is here:
https://gist.github.com/matthewhanson/6be66c97c828acd1d39d8cbb97a0981e

Issue #218, opened 2022-01-14
adopting stac-table and xstac


TomAugspurger:
Hi all,

I have a couple of small-ish packages to assist in creating STAC metadata for

1. Tabular data, like might be stored in parquet and loaded with pandas [`stac-table`][stac-table]
2. n-dimensional labeled data, like might be stored in NetCDF or Zarr and loaded with xarray [`xstac`][xstac]

I was wondering if stactools would be a good home for them.

Specifically, I'm proposing the addition of two new sub-modules / sub-packages to stactools:
Each of these subpackages would be either namespace packages or regular packages that require optional dependencies.
In other words, no new required dependecies for `stactools` itself, and no breaking API changes.

I think the new sub-packages should reflect the type of data they're tailored for. So I would propose something like

1. `stactools.table`
2. `stactools.datacube` (dunno about this name. But it matches the most prominent STAC extension used)

### Commit History

The easiest path forward is to just copy-paste the code from those libraries into the relevant submodules. Preserving the commit history is probably the responsible thing to do, but might be complicated. I'd recommend we give it one shot to make a pull request that preserves the commit history, and if that fails we can just copy the source itself.

### CLI

`xstac` includes [a CLI](https://github.com/TomAugspurger/xstac/blob/main/xstac/_generate.py#L93), based on `argparse`. stactools uses `click`. The `xstac` CLI should be rewritten to use click and updated to integrate into stactools

### Packaging

xstac and stac-table use a mix of setupools's declarative `setup.cfg` and pyproject.toml. I don't know how compatible those are with whatever we'll need, but they might need to be rewritten too.

### Tests

It seems like `stactools` is using `unittest` as a test runner. xstac and stac-table are using pytest.
pytest is really nice, and I would recommend it for any new project. Its test runner can handle modules that
use unittest's basic features, so we might be able to just switch out the runner. If that causes issues,
we'd need to either update stactools to use pytest, or xstac and stac-table to use unittest.

### Deprecating xstac and stac-table

We don't want people hanging around on old versions of xstac and stac-table. So once we have a release of stactools with these modules integrated, I'll make a release of xstac and stac-table that issues a warning at import time instructing users to upgrade.

[stac-table]: https://github.com/TomAugspurger/stac-table
[xstac]: https://github.com/TomAugspurger/xstac

## Alternatives

[...]

PR #307, opened 2022-06-09
Raster data footprint determination support

philvarner:
**Related Issue(s):**

- #85 

**Description:**

These changes add a new module `raster_footprint` that provides parameterized methods for generating data footprints from raster data. This includes a method to update the geometry of an Item based on the footprint determined from the data assets in the Item.

**PR checklist:**

- [X] Code is formatted (run `scripts/format`).
- [X] Code lints properly (run `scripts/lint`).
- [X] Tests pass (run `scripts/test`).
- [X] Documentation has been updated to reflect changes, if applicable.
- [X] Changes are added to the [CHANGELOG](../CHANGELOG.md).


pjhartzell:
Unsolicited opinions on the `data_extent` function:
- I'm not a pyproj expert, but after looking at the docs, I don't think the `accuracy` option involves densification. Looks more like coordinate computation accuracy guarantee.
- I think the check should be `arr != src.nodata` rather than against 0
- Checking against nodata will likely produce a lot of small polygons in certain rasters, so looking at `rasterio.features.sieve` might be something to look at.
- Interested whether the `scale` is beneficial in any way
- I think reprojecting the raster to wgs84 _before_ `rasterio.features.shapes` and `shapely.geometry.simplify` might be worth looking at. If you reproject to wgs84 after pulling out the shapes, you start having to think about densification. Might be easier to just avoid that.

matthewhanson:
Note that the issue for this is https://github.com/stac-utils/stactools/issues/85 where the initial draft code came from, and there's some examples of the code in action.

Some responses to @pjhartzell

- `[rasterio.features.sieve](https://github.com/rasterio/rasterio/blob/master/examples/sieve.py)` does look like it would be worth using instead and having a parameter the user can pass through to it.   
- Scale is likely not beneficial. The original idea was that you could generate a reasonable footprint by just reading an overview of the image rather than the whole array, but in some initial testing the results were too coarse.  Recommend additional testing though before ditching.
- Regarding densification, my experience is that the opposite problem tends to be the case, where the geometries are way too detailed.  `rasterio.features.shapes` can create highly detailed geometries and simplification is then required. The 2nd animated gif in issue #85 was simplified, the original polygon was massive.

matthewhanson:
I'll also add in here that while it's use case driven whether or not overestimating or underestimating the data geometry is preferable, I think the default method should slightly underestimate the data geometry.  

As a user I don't want to get matches against scenes that end up having only nodata under my polygon, whereas if I miss a scene because it happened to have a handful of pixels on the edge that intersected with my polygon....I just don't see that as much of a big deal.

matthewhanson:
Ah, gotcha @pjhartzell that makes perfect sense.

Knowing if the scaling factor is usable is even more important then, since if we can get away with using even a 1/2 overview then we greatly reduce computation.  Not a big deal if doing a handful of images, but over a million files this could amount to some sizable $.

pjhartzell:
Yeah, the computational aspect of this is a definite drawback.

TomAugspurger:
One meta-comment: this is assuming that you're cataloging raster data that can be read with rasterio, and wouldn't work for tabular or Zarr / NetCDF assets (https://github.com/stac-utils/stactools/issues/218). I think the names `data_extent` and `data_extents_for_data_assets`, or the namespace these functions are in, should reflect that it's for raster data.

philvarner:
@matthewhanson 

> I'll also add in here that while it's use case driven whether or not overestimating or underestimating the data geometry is preferable, I think the default method should slightly underestimate the data geometry.

I think we get a bit of each here -- using the convex hull for the data area will overestimate the data area if there are lots of no-data concavities. I'm not really sure what the general effect of densification/simplification is along the edges -- I think in many cases the over/under balances out in terms of total area, and there will be some edges that are a little over and some a little under.

philvarner:
@TomAugspurger 

> One meta-comment: this is assuming that you're cataloging raster data that can be read with rasterio, and wouldn't work for tabular or Zarr / NetCDF assets (#218). I think the names `data_extent` and `data_extents_for_data_assets`, or the namespace these functions are in, should reflect that it's for raster data.

renamed the module to "raster_footprint"

philvarner:
@matthewhanson @pjhartzell 

Thinking about this a little more, since I don't even understand how scale is supposed to be used, it should probably be documented better. Does someone want to explain it to me?

pjhartzell:
>sieve wasn't the right option here -- sieve will connect the small no data polygons, but then you just get larger complex polygons within footprint. Using the convex hull gives what I would consider a footprint. Future work might be to attempt to calculate the alpha shape (a sort of "concave hull") but I don't see a clear benefit in using the more-complex alpha shape vs. the simpler convex hull. I think this is only beneficial when there are large, highly concave no data areas that only intersect one side of an image, which aren't common. I could see this being useful for another custom property of an Item, but not for the Item geometry, where you generally want something simpler and less accurate than larger and more accurate.

I'm +1 on keeping things simple. I would push back a little in that if a `nodata` value is used to generate a polygon footprint, I think most users would expect the concave regions of `nodata` values to be excluded. So the use of a convex hull should be clear in the docs of the primary function(s), or maybe even in the function name. 

Agreed that sieve will connect the small nodata polygons into larger complex polygons within the footprint. But those larger complex polygons can then be simplified, no? The issue of holes, or inner polygons, will be an issue, but there might be some ways to remove inner polygons. But I'm open to leaving this to future work.

>I think I've balanced the densification / simplification concerns with the method parameters and the documentation around them. I explicitly put in the docs that users should emperically validate what the footprint looks like, and that they need to adjust the knobs to work for their particular data, and what those knobs mean. I think this is actually useful compared with a lot of the docs for the libraries i've used for this, which state what they do, but don't really explain how to configure things for different types of data.

I guess my concern on this is that projection distortion is not the same everywhere. So a savvy user will pick an area with the worst distortion and tune things there. Or just use a very high densification factor (easier). But if just using a very high densification factor will do the trick, is there any reason we can't do this (with a bit more intelligence that just a very high value) for the user? And avoid users having to empirically tune the densification? Again, perhaps a future improvement.

Speaking of densification, does densification by distance rather than by factor make sense? When densifying by factor, you can end up with a lot of very closely spaced points where the original polygon has very close vertices, and vice versa where the original vertices are far apart.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,313 @@
...
+     Args:
+        polygon (Polygon): The input Polygon
+        crs (CRS): The CRS of the input Polygon
+        densification_factor (Optional[int]): The factor by which to increase point density
+            within the polygon. A factor of 2 would double the density of points (placing one
+            new point between each existing pair of points), a factor of 3 would place two points
+            between each point, etc.
+        precision (int): The number of decimal places to include in the coordinates for the
+            reprojected geometry. Defaults to 3 decimal places.
+        simplify_tolerance (Optional[float]): The minimum distance to maintain between points,
```
pjhartzell, line 293: The tolerance in shapely's simplify is not the minimum distance between points. It's more like a maximum orthogonal distance between each of the original shape's vertices and the lines connecting the simplified shape's vertices.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,313 @@
...
+            in degrees.
+        preserve_topology (bool): Preserve topology during simplification.
```
pjhartzell, line 295: I believe preserving topology is only relevant for MultiPolygons (prevent overlaps and gaps between Polygons in a MultiPolygon when simplifying). As it stands, MultiPolygons are not used (a single convex hull Polygon is being simplified), so this option has no effect.

philvarner:
The convexity and inclusion of nodata should be sufficiently documented. Let me know if you think it needs more.
 
> I guess my concern on this is that projection distortion is not the same everywhere. So a savvy user will pick an area with the worst distortion and tune things there. Or just use a very high densification factor (easier). But if just using a very high densification factor will do the trick, is there any reason we can't do this (with a bit more intelligence that just a very high value) for the user? And avoid users having to empirically tune the densification? Again, perhaps a future improvement.

This is a good point -- we may want to just put the densification factor to default to 10. This makes reprojection 10x slower. The additional points get cleaned up by the simplification, so the resulting geometry isn't necessarily 10x larger. 
 
> Speaking of densification, does densification by distance rather than by factor make sense? When densifying by factor, you can end up with a lot of very closely spaced points where the original polygon has very close vertices, and vice versa where the original vertices are far apart.

I think it would - we'd have to write our own code for this, as shapely only does equidistant densification.

src/stactools/core/utils/raster_footprint.py
philvarner: fixed

pjhartzell:
>we may want to just put the densification factor to default to 10. This makes reprojection 10x slower.

Point reprojection should be pretty cheap, though. At least compared to reprojecting an image. So I think it'd be OK to add more points.

> as shapely only does equidistant densification

I think a custom function (`_densify`) is currently being used for the densification. Here's a shameless [link](https://github.com/pjhartzell/reproject-geometry/blob/fb0b0bb4b25497da513550150bb955b702ba4a00/src/reproject/reproject.py#L205) to a by-distance densification function.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+        Polygon: the reprojected Polygon
+    """
+    if densification_factor is not None:
+        polygon = Polygon(_densify(polygon.exterior.coords, densification_factor))
+
+    polygon = shape(transform_geom(crs, "EPSG:4326", polygon, precision=precision))
+
+    if simplify_tolerance is not None:
+        polygon = polygon.simplify(
+            tolerance=simplify_tolerance, preserve_topology=False
```
pjhartzell, line 295: I deleted a prior comment where I recommended setting `preserve_topology` to `False` since my understanding of topology preservation only applies to multiple poygons (prevent overlaps and gaps from adjacent polygons). I deleted the comment because a bit more research revealed that some simplification algorithms do produce different polygon depending on a `preserve_topology`-like setting, and I don't know the characteristics of shapely's algorithms. Perhaps just remove the explicit setting and let shapely's default of `True` for `preserve_topology` to hold.

pjhartzell (approved):
Other than a suggestion to remove the explicit `preserve_topology` setting in `densify_reproject_simplify`, I think this is good to go. We can look at handling concavity and perhaps a refined or automatic "densification" method going forward.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+            the polygon.
+            A factor of 2 would double the density of points (placing one new point between each
+            existing pair
+            of points), a factor of 3 would place two points between each point, etc.
+        simplify_tolerance (Optional[float]): All points in the simplified object will be within
+            the tolerance distance of the original geometry, in degrees.
+
+    Returns:
+        Iterator[Tuple[str, Dict[str, Any]]]: Iterator of the data extent as a geojson dict
+            for each asset.
```
gadomski, line 69:
```suggestion
       Optional[pystac.Item]: The modified item, or ``None`` if the item was unmodified.
```

It feels a little awkward to have a "partially-fluent" interface, where you can't rely on the item being returned. IMO we should either
- always return the item, or
- return a boolean indicating whether the item was modified

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+    asset_names: Optional[List[str]] = None,
+    scale: int = DEFAULT_SCALE,
+    precision: int = DEFAULT_PRECISION,
+    densification_factor: Optional[int] = None,
+    simplify_tolerance: Optional[float] = None,
+    no_data: Optional[int] = None,
+) -> Iterator[Tuple[str, Dict[str, Any]]]:
+    """
+    Accepts an Item and an optional list of asset names within that item, and produces
+    an iterator over the same asset names (if they exist and have the role "data") and
```
gadomski, line 103: While I agree that `"data"` is a good default role, this should be configurable -- there could be providers that use something other than `data`.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+        Iterator[Tuple[str, Dict[str, Any]]]: Iterator of the data extent as a geojson dict
+            for each asset.
+    """
+    for name, asset in item.assets.items():
+        if (
+            asset.roles
+            and "data" in asset.roles
+            and (asset_names is None or name in asset_names)
+        ):
+            href = make_absolute_href(asset.href, item.get_self_href())
```
gadomski, line 136:
```suggestion
            href = asset.get_absolute_href()
            if href is none:
                logger.error(f"Could not determine extent for asset '{name}'")
                continue
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+    href: str,
+    *,
+    scale: int = DEFAULT_SCALE,
+    precision: int = DEFAULT_PRECISION,
+    densification_factor: Optional[int] = None,
+    simplify_tolerance: Optional[float] = None,
+    no_data: Optional[int] = None,
+) -> Optional[Dict[str, Any]]:
+    """
+        Produces a data footprint from an asset.
```
gadomski, line 189:
```suggestion
        Produces a data footprint from the href of a raster file.
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+
+        See :py:meth:`update_geometry_from_asset_footprint` for more details about densification
+        and simplification.
+
+     Args:
+        href (str):
```
gadomski, line 195: Needs docstring.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,301 @@
...
+        src.crs,
+        densification_factor=densification_factor,
+        precision=precision,
+        simplify_tolerance=simplify_tolerance,
+    )
+
+    import json
+
+    print(json.dumps(mapping(polygon)))
+
```
gadomski, line 254: Leftover debugging?

tests/core/utils/test_raster_footprint.py
```diff
@@ -0,0 +1,354 @@
...
+    item = pystac.read_file(
+        test_data.get_path(
+            "data-files/raster_footprint/MCD43A4.A2001055.h25v06.006.2016113010159.json"
+        )
+    )
+
+    item = update_geometry_from_asset_footprint(
+        item, asset_names=["B01"], densification_factor=10
+    )
+    # print(json.dumps(item.geometry))
```
gadomski, line 46: Comment should be removed.

gadomski (changes requested):

src/stactools/core/utils/raster_footprint.py
philvarner: I'd removed this but forgot to push it

src/stactools/core/utils/raster_footprint.py
philvarner: yeah, I agree. changed.

src/stactools/core/utils/raster_footprint.py
philvarner: so this is odd, because the GEOS documentation says that it doesn't apply to polygons, but I got different results with True and False for only one of our footprints. The test that checks precision setting simplified to a polygon that had a line in it like, but with False it removed that line, which is better, but surprising because it wasn't supposed to affect polygons! Maybe this is an outlier case? 

|-----|
|-----|
|-----|
|--------------

src/stactools/core/utils/raster_footprint.py
philvarner: Ok, will add this.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,296 @@
...
+    in a reprojected polygon that does not accurately represent the data footprint. For
+    example, a MODIS asset represented by a rectangular 5 point Polygon in a sinusoidal
+    projection will reproject to a parallelogram in EPSG 4326, when it would be more
+    accurately represented by a polygon with two parallel sides and two curved sides.
+    The difference between these representations is even greater the further away from
+    the meridian and equator the asset is located.
+
+    After reprojection to EPSG 4326, the footprint may have more points than desired.
+    This can be simplified to a polygon with fewer points that maintain a minimum distance
+    between each point.
```
pjhartzell, line 50: Update to match corrected description of simplification (as in the `simplify_tolerance` description in Args below).

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,296 @@
...
+
+    Args:
+        item (Item): The PySTAC Item to extend.
+        asset_names (Optional[List[str]]): The names of the assets for which to extract footprints.
+        scale (int): A factor to scale out the shape to. Defaults to 1.
```
pjhartzell, line 55: Meh. I'd be ok with removing the scale option. But should probably update the description to clarify that it can be used to scale "out" or "in".

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,296 @@
...
+            scale=scale,
+            precision=precision,
+            densification_factor=densification_factor,
+            simplify_tolerance=simplify_tolerance,
+            no_data=no_data,
+        ),
+        None,
+    )
+    if asset_name_and_extent is not None:
+        _, extent = asset_name_and_extent
```
pjhartzell, line 83: I'm not clear on this. What is the purpose of allowing multiple asset_names if only the first one (returned by `next`) is used?

philvarner: in case your items have different assets, so they're in decreasing preference, e.g., "blue_10m", "blue_60m", etc.

src/stactools/core/utils/raster_footprint.py
philvarner: I might just remove it now that I'm not so eager to just merge it

src/stactools/core/utils/raster_footprint.py
pjhartzell: So if the footprint fails for the first asset, it tries on the second asset, and so on?

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,296 @@
...
+            between each point, etc.
+        simplify_tolerance (Optional[float]): All points in the simplified object will be within
+            the tolerance distance of the original geometry, in degrees.
+        no_data(Optional[int]): explicitly set the no data value if not in image metadata
+    Returns:
+        Iterator[Tuple[str, Dict[str, Any]]]: Iterator of the data extent as a geojson dict
+            for each asset.
+    """
+    for name, asset in item.assets.items():
+        if asset.roles and (asset_names is None or name in asset_names):
```
pjhartzell, line 129: What is the reason for the requirement for `roles` to exist on the asset?

src/stactools/core/utils/raster_footprint.py
philvarner: yes

philvarner: or more likely than failing is that it doesn't exist

src/stactools/core/utils/raster_footprint.py
philvarner: oh, that's unnecessary now -- previously we were checking that the role was data

philvarner: I'll fix that up, and I also see there's a mention in the documenation to remove

gadomski (approved):

src/stactools/core/utils/raster_footprint.py
pjhartzell: IIUC, simplification does not impart a minimum distance between points; it looks at the distance between a point and the new geometry should that point be removed. See [here].(https://shapely.readthedocs.io/en/stable/manual.html#object.simplify)

The sentence should read something like, "This can be simplified to a polygon with fewer points that maintain a minimum distance _to the original geometry_."

gadomski:
@philvarner @pjhartzell I've been playing with this branch a bit this morning, and when I checked it against the landsat test data, I was surprised to find that the raster footprint did not include all pixels:
landsat-raster-footprint

Is this expected behavior?

gadomski:
@philvarner @pjhartzell I've made a couple of additions/changes to this PR:
1. Added a command line interface, so you can now do `stac update-geometry item.json`
2. Reduced the size of the landsat test file -- it was ~95MB, so I cropped it down to ~3MB

I re-requested your review @pjhartzell, if you wouldn't mind looking at the command line interface in particular since no one else has put eyes on that.

NB I've done a rebase of this branch (which is how I found the large file) so you'll have to force reset any local branches.

pjhartzell:
@gadomski Still working through the review, but in answer to your question, the reason all the pixels are not included is because the `DEFAULT_PRECISION = 3` is very low for geodetic coordinates. Probably a leftover from thinking about things in meters. As a _srrveyor_, my default precision for geodetic coordinates is 8, which is roughly a millimeter at the equator. Rule of thumb for geodetic precision is:
- 0 = 100,000m
- 1 = 10,000m
- 2 = 1,000m
- 3 = 100m
- 4 = 10m
- 5 = 1m
- 6 = 0.1m

and so on. Here's a snapshot of the footprint with a precision of 3 (red) and 8 (green). I did not apply any simplification, so a precision of 8 wraps around each pixel:
image

Of course this leads to a very large very large geometry coordinate list. If we keep a precision equal to 8 and set a simplification tolerance equal to roughly twice the pixel size (30mx2 = 60m --> 0.0006 degrees for this landsat band), we get something closer to what we expect (blue line):
image

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,284 @@
...
+from pystac import Item
+from rasterio.crs import CRS
+from rasterio.warp import transform_geom
+from shapely.geometry import mapping, shape
+from shapely.geometry.multipolygon import MultiPolygon
+from shapely.geometry.polygon import Polygon
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PRECISION = 3
```
pjhartzell, line 18: As mentioned in a comment in the general conversation thread, this should probably be changed to 7 (centimeter) or 8 (millimeter).

pjhartzell (changes requested):

tests/data-files/raster_footprint/LC08_L1TP_198029_20220331_20220406_02_T1_B2.json
```diff
@@ -0,0 +1,346 @@
+{
+  "assets": {
+    "B2": {
+      "gsd": 500,
+      "href": "./LC08_L1TP_198029_20220331_20220406_02_T1_B2.TIF",
```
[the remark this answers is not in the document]
gadomski: Yeah, I discovered that as well. I don't love it, but it works?

src/stactools/cli/commands/update_geometry.py
```diff
@@ -0,0 +1,74 @@
...
+        item_path: str,
+        asset_names: List[str],
+        precision: int,
+        densification_factor: Optional[int],
+        simplify_tolerance: Optional[float],
+        no_data: Optional[int],
+    ) -> None:
+        item = Item.from_file(item_path)
+        if not asset_names:
+            optional_asset_names = None
```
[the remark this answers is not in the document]
gadomski: I chose an alternate fix, I made `asset_names` just a `List[str]` (instead of an `Optional[List[str]]`): https://github.com/stac-utils/stactools/pull/307/commits/d39aa095f0677d0bef9a9e8afa8ecfc41c01982c

tests/data-files/raster_footprint/LC08_L1TP_198029_20220331_20220406_02_T1_B2.json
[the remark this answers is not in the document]
pjhartzell: Yep.

src/stactools/core/utils/raster_footprint.py
pjhartzell: The geometry is always going to to be WGS84 geodetic coords, so I think we can just use a single value?

gadomski: Ah, I wasn't sure if the precision was used before or after reprojection, but if it's after, then I agree, will change to centimeter (7).

src/stactools/core/utils/raster_footprint.py
```diff
@@ -0,0 +1,286 @@
...
+    The difference between these representations is even greater the further away from
+    the meridian and equator the asset is located.
+
+    After reprojection to EPSG 4326, the footprint may have more points than desired.
+    This can be simplified to a polygon with fewer points that maintain a minimum distance
+    to the original geometry.
+
+    Args:
+        item (Item): The PySTAC Item to extend.
+        asset_names (List[str]): The names of the assets for which to extract footprints. If the
```
pjhartzell, line 52: This, to me, is not clear. It implies that footprints will be extracted for all assets in the list. Which then is not clear how the multiple footprints from the multiple assets would be combined. But this is not the case; IIUC, the first successful footprint from the list of assets is used - all remaining assets are ignored. Likewise, if no asset list is provided, then the footprint generated from first successful asset in the Item is used. Is that right?

pjhartzell (commented):
Ran the CLI on some files. Works well. Looks like the updated precision broke the tests, though. 🙁

src/stactools/core/utils/raster_footprint.py
gadomski: > the first successful footprint from the list of assets is used - all remaining assets are ignored

Correct. Which brings up the interesting question on whether asset order is preserved when loading into PySTAC. If not, results could be non-deterministic.

> if no asset list is provided, then the footprint generated from first successful asset in the Item is used. Is that right?

Correct.

gadomski: I'll revise the text to be more clear.

pjhartzell (approved):
Looks good.

PR #317, opened 2022-06-16
Antimeridian fixes now allow multipolygons

### LINKED ISSUE #316: Handle MultiPolygon Item geometries in the antimeridian `fix_item` utility function
[`fix_item`](https://github.com/stac-utils/stactools/blob/1077ccd58cafc4dd392abddefc490af2ee64c1ea/src/stactools/core/utils/antimeridian.py#L25) will only split or normalize a Polygon Item geometry (raises an error otherwise). This precludes, for example, the ability to create a single, normalized Polygon if an Item's geometry is already split along the Antimeridian and therefore consists of a MultiPolygon.

The Landsat stactools package contains an [example workaround](https://github.com/stactools-packages/landsat/blob/cc6992b6e7655bf30265d79c1b87746fc25302d3/src/stactools/landsat/utils.py#L233) that merges a MultiPolygon Item geometry to single Polygon before calling `fix_item` to generate a single, normalized Polgyon geometry. However, this workaround and the brief discussion above assumes the MultiPolygon is split along the Antimeridian. Cases where a MultiPolygon can not be merged into a single Polygon would also need to be handled.

gadomski:
**Related Issue(s):**
- Closes #316 

**Description:**
Extend the antimeridian module to allow multipolygons. The `fix_item` function checks whether the geometry is a polygon or multi-polygon, and reacts accordingly. MultiPolygons are split/normalized via new functions to keep the existing functionality untouched, since they require slightly different logic to work.

**PR checklist:**

### DIFF
--- a/src/stactools/core/utils/antimeridian.py
+++ b/src/stactools/core/utils/antimeridian.py
@@ -24,7 +24,9 @@ class Strategy(Enum):
 
 def fix_item(item: Item, strategy: Strategy) -> Item:
     """Modifies an item in-place to deal with antimeridian issues.
-    If the item's geometry is not a `Polygon`, raises a `ValueError`.
+
+    If the item's geometry is not a :py:class:`Polygon`` or a
+    :py:class:`MultiPolygon`, raises a :py:class:`ValueError`.
 
     Args:
         item (pystac.Item): The item to be modified.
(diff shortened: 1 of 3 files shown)


pjhartzell (commented):
While this does handle MultiPolygons, it doesn't quite address the issue in the manner I was originally thinking. But I think that is OK for the most part. 

Setting aside the issue of merging a MultiPolygon that is already split along the antimeridian (that is a special case that should likely stay in the domain of the user to handle before calling `fix_item`), I'm thinking about what a user expects when asking to normalize a MultiPolygon where some Polygons are west of, some are east of, and some span the antimeridian. Do they expect the x-coordinates of all Polygons in the MultiPolygon to be of the same sign? Or do they only expect each Polygon to be consistent in itself? As it stands, each Polygon is made consistent in itself, which means a normalized MultiPolygon can contain Polygons with different x-coordinate signs. This is sufficient to avoid individual Polygons stretching across the globe. But it will make for bounding boxes that stretch across the globe.

It is a much simpler task for a user to change the signs of consistent Polygons after calling `fix_item` or `normalize_multipolygon`. But the bounding box spanning the globe is not ideal.

If nothing else, the effect on individual Polygons in a MultiPolygon (not all Polygons will have the same x-coordinate sign, but each Polygon will have a consistent x-coordinate sign) should be made more clear in the `normalize_multipolygon`'s documentation string.

gadomski:
>This is sufficient to avoid individual Polygons stretching across the globe. But it will make for bounding boxes that stretch across the globe.
> It is a much simpler task for a user to change the signs of consistent Polygons after calling fix_item or normalize_multipolygon. But the bounding box spanning the globe is not ideal.

This makes a lot of sense and won't be (too hard) of an improvement. I'll add this, thanks.

src/stactools/core/utils/antimeridian.py
```diff
@@ -149,3 +191,36 @@ def normalize(polygon: Polygon) -> Optional[Polygon]:
         return shapely.affinity.translate(polygon, xoff=+360)
     else:
         return polygon
+
+
+def normalize_multipolygon(multi_polygon: MultiPolygon) -> Optional[MultiPolygon]:
+    """'Normalizes' a WGS84 lat/lon multi polygon, or returns None if no changes were made.
+
+    This converts the multi-polygon's x coordinates to all be the same sign, even if
+    the polygon crosses the antimeridian.
```
pjhartzell, line 200: Maybe update the explanation to make the outcome clear. Something like:
>For each polygon in the multi-polygon, this converts the x coordinates to all be the same sign, even if the polygon crosses the antimeridian. Although the x coordinate sign within each polygon will be made the same, the sign may differ between polygons depending on their position relative to the antimeridian.

pjhartzell (approved):
Just one suggestion on clarifying the result of `normalize_multipolygon`.

src/stactools/core/utils/antimeridian.py
gadomski: Updated w/ your text.

geomatician (approved):

Issue #224, opened 2022-01-27
Extend `cogify` to extract one band from a multi-band source


gadomski:
We decided to keep `cogify` in #222, but it currently only works for single-band data sources (e.g. as discussed here: https://github.com/pjhartzell/nclimgrid/pull/2#discussion_r791897187). We should extend `cogify` to work for creating a single-band COG from multi-band sources.

PR #318, opened 2022-06-17
Extend cogify

geomatician:

**Related Issue(s):**

- https://github.com/stac-utils/stactools/issues/224

**Description:**
Exended `cogify` to:
- extract single-band COGs from multi-band images
- export subdatasets from HDF files as separate COGs

**PR checklist:**


src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
+
+def cogify(
+    infile: str,
+    outfile: str,
+    band: Optional[int] = None,
+    profile: Optional[Dict[str, Any]] = None,
+) -> None:
+    """Creates a Cloud-Optimized GeoTIFF (COG) from a GDAL-readable file
+    without subdatasets (non-HDF files). A band number can optionally be
+    provided to extract a single band from a multiband file.
```
gadomski, line 35:
```suggestion
    """Creates a Cloud-Optimized GeoTIFF (COG) from a GDAL-readable file.
    
    A band number can optionally be provided to extract a single band from a
    multiband file. To create COGs from subdatasets, use :py:meth:`stactools.core.utils.convert.cogify_subdatasets`.
```

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
 
     Uses :py:meth:`rasterio.shutil.copy`.
 
     Args:
         infile (str): The input file.
         outfile (str): The output COG to be written.
+        band (Optional[int]): The band number in the input file to extract.
```
gadomski, line 42:
```suggestion
        band (Optional[int]):
            The band number in the input file to extract. If not provided, a multi-band COG will be created.
```

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
-            "GDAL's COG driver is not enabled, make sure you're using GDAL >= 3.1"
-        )
-    destination_profile = DEFAULT_PROFILE.copy()
+    check_gdal_driver()
+
+    src = rasterio.open(infile)
+    dest_profile = DEFAULT_PROFILE.copy()
+    dest_profile.update(
+        {
+            "dtype": rasterio.uint8,
```
gadomski, line 55: Not all output datasets will have a `unit8` datatype.

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
+
+def cogify_subdataset(infile: str, outfile: str) -> None:
+    """Exports a Cloud-Optimized GeoTIFF (COG) from a subdataset within an HDF file.
+    Uses :py:meth:`rasterio.shutil.copy`.
+    Args:
+        infile (str): The input file.
+        outfile (str): The output COG to be written.
+    """
+    check_gdal_driver()
+    destination_profile = DEFAULT_PROFILE.copy()
```
gadomski, line 86: Is this function necessary? it looks very similar to `cogify` with `bands = None`.

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
     rasterio.shutil.copy(infile, outfile, **destination_profile)
+
+
+def list_subdataset(infile: str, outdir: str) -> Tuple[List[str], List[str]]:
```
gadomski, line 90:
```suggestion
def cogify_subdatasets(infile: str, outdir: str) -> Tuple[List[str], List[str]]:
```

Can we also add an optional argument that is a list of subdataset names that can be used to subset the output datasets? E.g.:

```python
def cogify_subdatasets(infile: str, outdir: str, subdataset_names: Optional[List[str]] = None) -> ...
```

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
+    """Generates lists of output paths and subdataset names for COGs created from an HDF file.
+    This is then used to call the cogify_subdataset() function to export out COG files.
+    Args:
+        infile (str): The input HDF file
+        outdir (str): The output directory where the HDF files will be created
+    Returns:
+        Tuple[List[str], List[str]]: A two tuple (paths, names):
+            - The first element is a list of the output tiff paths
+            - The second element is a list of subdataset names
+    """
```
gadomski, line 100:
```suggestion
    """Creates Cloud-Optimized GeoTIFFs for all subdatasets in a multi-dataset raster file.

   The created files will be named the same as the source file, with a ``_SUBDATASET`` suffix. E.g. if the source file is named ``foo.hdf`` and the subdataset is named ``bar``, the output COG will be named ``foo_bar.tif``.

    Args:
        infile (str): The input file containing subdatasets.
        outdir (str): The output directory where the COGs will be created.
    Returns:
        Tuple[List[str], List[str]]:
            A two tuple (paths, names):
                - The first element is a list of the output COG paths
                - The second element is a list of subdataset names
    """
```

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,102 @@
...
+            parts = subdataset.split(":")
+            subdataset_name = parts[-1]
+            sanitized_subdataset_name = subdataset_name.replace(" ", "_").replace(
+                "/", "_"
+            )
+            subdataset_names.append(sanitized_subdataset_name)
+            file_name = f"{base_file_name}_{sanitized_subdataset_name}.tif"
+            outfile = os.path.join(outdir, file_name)
+            cogify_subdataset(subdataset, outfile)
+            paths.append(outfile)
```
gadomski, line 117: We should add some logic to make sure the subdataset can be a valid raster -- e.g. sometimes there are single variable subdatasets that don't make sense to convert. See https://github.com/stactools-packages/viirs/blob/ad062e9d84dcfef05009f78453b2c6fa849fdab7/src/stactools/viirs/cog.py#L56-L61 for an example of this type of check.

gadomski (changes requested):

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,101 @@
...
+    """
+
+    with rasterio.open(infile) as dataset:
+        assert_cog_driver_is_enabled()
+        subdatasets = cast(List[str], dataset.subdatasets)
+        base_file_name = os.path.splitext(os.path.basename(infile))[0]
+        paths = []
+        names = []
+        for subdataset in subdatasets:
+            subd = rasterio.open(subdataset)
```
gadomski, line 101: I'm not sure if we need to re-open the subdataset with rasterio. However, I can't check, because there's no unit test for the subdataset case.

src/stactools/core/utils/convert.py
```diff
@@ -16,25 +17,101 @@
...
+        rasterio.shutil.copy(infile, outfile, **dest_profile)
+
+
+def cogify_subdatasets(
+    infile: str, outdir: str, subdataset_names: Optional[List[str]] = None
+) -> Tuple[List[str], List[str]]:
+    """Creates Cloud-Optimized GeoTIFFs for all subdatasets in a multi-dataset raster file.
+    The created files will be named the same as the source file, with a ``_SUBDATASET`` suffix.
+    E.g. if the source file is named ``foo.hdf`` and the subdataset is named ``bar``, the output
+    COG will be named ``foo_bar.tif``.
```
gadomski, line 83: Probably worth adding a note about how only 2D subdatasets are supported, not 3D.

gadomski (changes requested):

src/stactools/core/utils/convert.py
geomatician: I believe you do need to reopen the subdataset, since `dataset.subdatasets` only returns a list of strings containing the subdataset names. To loop through subdatasets and export each out as a separate file, you need to open each subdataset. I added a unit test for this function. 
Again, happy to be informed otherwise.

gadomski (changes requested):

src/stactools/core/utils/convert.py
```diff
@@ -101,20 +101,21 @@ def cogify_subdatasets(
...
+                sanitized_subdataset_name = subdataset_name.replace(" ", "_").replace(
+                    "/", "_"
+                )
+                names.append(sanitized_subdataset_name)
+                file_name = f"{base_file_name}_{sanitized_subdataset_name}.tif"
+                outfile = os.path.join(outdir, file_name)
+                destination_profile = DEFAULT_PROFILE.copy()
+                rasterio.shutil.copy(subdataset, outfile, **destination_profile)
+                paths.append(outfile)
+                subd.close()
```
gadomski, line 120: Don't need to close, the `with` statement will take care of that for us.

gadomski (changes requested):

gadomski (approved):

PR #321, opened 2022-06-24
rename addraster command to add-raster

philvarner:

**Related Issue(s):**

- https://github.com/stac-utils/stactools/issues/322

**Description:**

renamed `addraster` command to `add-raster`.  This aligns the naming convention of `create-item`, and provides a consistent pattern going forward. 

Since this is a minor version release <1.0, I do not think adding the new command and deprecating the existing one is necessary, and the failure because of the now non-existant command will be obvious.

**PR checklist:**

- [X] Code is formatted (run `scripts/format`).
- [X] Code lints properly (run `scripts/lint`).
- [X] Tests pass (run `scripts/test`).
- [X] Documentation has been updated to reflect changes, if applicable.
- [X] Changes are added to the [CHANGELOG](../CHANGELOG.md).


gadomski (approved):

gadomski:
> Since this is a minor version release <1.0, I do not think adding the new command and deprecating the existing one is necessary, and the failure because of the now non-existant command will be obvious.

Agreed, and brings up the question of whether we consider the command line interface part of the "public API" for semver reasons. Thoughts?

philvarner:
I would say it is part of the public api, so we only *must* maintain compatibility within the patch releases for <1.0

gadomski:
> I would say it is part of the public api, so we only must maintain compatibility within the patch releases for <1.0

👍🏽

PR #331, opened 2022-07-19
Add utility to ignore NotGeoreferencedWarning

gadomski:
**Related Issue(s):**
None

**Description:** Used for digging into subdatasets of, e.g., hdf files. Previous art: https://github.com/stactools-packages/viirs/blob/455ca9600c539999a882f5bfc1f20871103470a0/src/stactools/viirs/utils.py#L19-L26

**PR checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Specify installation channel to use for all conda packages to avoid incompatibility ([#301](https://github.com/stac-utils/stactools/pull/301))
 - Allow MultiPolygons when fixing antimeridian issues ([#317](https://github.com/stac-utils/stactools/pull/317))
 - Conda package, via [conda-forge](https://anaconda.org/conda-forge/stactools) ([#324](https://github.com/stac-utils/stactools/pull/324))
+- Context manager to ignore rasterio's NotGeoreferencedWarning ([#331](https://github.com/stac-utils/stactools/pull/331))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


src/stactools/core/utils/convert.py
```diff
@@ -95,26 +95,27 @@ def cogify_subdatasets(
...
-                    "/", "_"
-                )
-                names.append(sanitized_subdataset_name)
-                file_name = f"{base_file_name}_{sanitized_subdataset_name}.tif"
-                outfile = os.path.join(outdir, file_name)
-                destination_profile = DEFAULT_PROFILE.copy()
-                rasterio.shutil.copy(subdataset, outfile, **destination_profile)
-                paths.append(outfile)
-        return (paths, names)
+    with utils.ignore_not_georeferenced():
```
pjhartzell, line 98: I like having the ignore context utility available via stactools. But I'm shaky on whether to apply it here. IIRC, this warning will get thrown in some cases even when rasterio can find georeferencing information for a subdataset of interest.

However, applying it here will also suppress georeferencing warnings when rasterio can not find subdataset georeferencing information (when you want to see the warning). This is the case for VIIRS, but the warning is suppressed since [georefererencing information is manually being added](https://github.com/stactools-packages/viirs/blob/455ca9600c539999a882f5bfc1f20871103470a0/src/stactools/viirs/cog.py#L115-L116) to the output COGs.

The downside of not applying it here is that a user will have to manually decorate or wrap their call to this function with the new `ignore_not_georeferenced` context manager. But I think it's better to let the user know when something is _potentially_ not georeferenced and allow them to wrap the call with the ignore context if they decide it is not relevant.

gadomski: Ok, yeah that makes sense. I'll remove this usage, and just decorate the test more (this was mostly motivated by me wanting the tests to be quiet 🤫 ).

pjhartzell (approved):

PR #355, opened 2022-09-14
Multi band raster footprints

gadomski:
**Description:**
Allow computation of raster footprints from multiple (or all) bands in an asset.

This is supported by adding a `bands` argument to a few raster footprint functions. This is a list of band numbers; if it is an empty list, all bands are used.

The footprint computation procedure is changed to use a pre-constructed byte array of zeros. Each band is read and its non-no-data pixels are set to `1` in the byte array. This slightly increases the memory requirements for the operation (one raster-sized byte array) but I don't think this _should_ be a problem.

Multi-band support is added to the `stac update-geometry` command via a `--bands` option, which takes a comma-delimited list or the special string `all`.

This PR includes a sidecar fix for the create-items command, which was producing incorrect JSON.

**PR checklist:**

## Motivation

Many electro-optical systems, e.g. ASTER, produce assets which have slightly different data footprints per band. The previous implementation of raster footprints only used the first band. Here's a GIF to demonstrate the fix:

multi-band-raster


src/stactools/core/utils/raster_footprint.py
```diff
@@ -208,23 +225,32 @@ def data_footprint(
...
+            if np.isnan(no_data):
+                data[~np.isnan(band_data)] = 1
+            else:
+                data[np.where(band_data != no_data)] = 1
         else:
-            arr[numpy.where(arr != no_data)] = data_val
-    else:
-        arr.fill(data_val)
+            data.fill(1)
+            break
```
pjhartzell, line 246: Why the `break` here?

pjhartzell (commented):
Just one question

src/stactools/core/utils/raster_footprint.py
gadomski: If all the pixels are `1`, then the data array can't be modified in any future loop (we're OR-ing the pixels, not AND-ing them).

Now that I look at it though, we could just skip the loop all together since this branch only hits if there is no nodata value, it doesn't care about the band data. I'll fix.

pjhartzell (approved):

PR #396, opened 2023-02-03
Move raster footprint functionality into a class

pjhartzell:
## Updated Description

**Related Issue(s):**
None.

**Description:**
The raster footprint utility often produces incorrect geometry polygons when the data spans the edge of the sinusoidal projection used by MODIS and VIIRS. A fix requires an alternative densification approach and a custom reprojection function. There is currently no convenient way to override these steps in the raster footprint utility since they are: 1) bundled into a single function (`densify_reproject_simplify`) and 2) that function exists at the end of a chain of function calls. 

This PR replicates the functionality of the raster footprint utility in a class (named `RasterFootprint`) with an eye toward subclassing for selective method overrides. In addition to breaking out individual methods for the densify, reproject, and simplify steps, alternative constructors are provided for footprint creation directly from an href or from a file opened with rasterio (a rasterio [DatasetReader](https://rasterio.readthedocs.io/en/latest/api/rasterio.io.html#rasterio.io.DatasetReader) object). An alternative densification function is also provided; you can now choose to densify by factor or by distance.

The existing raster footprint functions are retained, but marked as deprecated. Their functionality is fully contained in the new `RasterFootprint` class and the functions themselves now rely on `RasterFootprint`.

**PR checklist:**

## Original Description (for context)

**Related Issue(s):**
None.

**Description:**
The raster footprint utility often produces incorrect and globe-spanning geometry polygons when the data is on the edge of the sinusoidal projection used by MODIS and VIIRS.
- The nature of the sinusoidal projection precludes data that spans the antimeridian, so the globe-spanning polygons are a symptom of bad vertices.
- It appears that vertices very close to the projection edge are occasionally projected beyond the antimeridian. Some experimenting with the basic sinusoidal projection equations that are part of this PR indicate a high sensitivity to small changes in the coordinates and/or projection parameters at the projection edges.

This PR provides a fix to create correct raster footprints at the edges of the sinusoidal projection.

**Approach:**
- Densely infill the footprint polygon (using the ground pixel size as the densify interval) while still in grid coordinates.
- Project the densified grid polygon vertices to spherical lon/lat (using equations provided by [a MODIS user guide](https://modis-fire.umd.edu/files/MODIS_C6_BA_User_Guide_1.2.pdf)) and throw out any coordinates with an `abs(longitude)` value greater than `180`. This should be able to be done without the custom projection equations, but I am unable to force `proj` to not wrap the longitude coordinates.
- By densifying at the ground distance of a single pixel, we should be within at least one pixel of the antimeridian (should the furthermost pixel be thrown out due to reprojection error/instability).

**Notes:**
- I converted the existing raster geometry code to a base class, which is then subclassed to recreate the original functionality (the "Default" subclass) and new functionality that handles sinusoidal data (the "Sinusoidal" subclass). This is fine, but it does require some gymnastics to retain the existing function signatures.
- There is a new, but unused, `sinusoidal_pixel_to_grid` function in `projection.py`. It will be used by the MODIS stactools-package and it seemed natural to place it alongside the new `sinusoidal_grid_to_lonlat` function.
- **I changed the name of an underscored/private function:** `_densify()` to `_densify_by_factor()`. The name parallels the new `_densify_by_distance()` function used by the sinusoidal class.

**Example:**
Current, incorrect behavior (pink) for an h01v11 MODIS tile (MCD15A2H.A2022025.h01v11.061):
image

Proposed, correct behavior (green) via this PR for the same MODIS tile:
image


src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+                        yield name, extent
+                    else:
+                        logger.error(f"Could not determine extent for asset '{name}'")
+
+    def data_footprint(self, href: str) -> Optional[Dict[str, Any]]:
+        with rasterio.open(href) as src:
+            self.transform = src.transform
+            self.shape = src.shape
+            if self.crs is None:
+                self.crs = src.crs
```
gadomski, line 148: I'm a little wary of mutating the underlying `RasterFootprint` object here, unless we need to. What do you think about:

```suggestion
            if self.crs is None:
               crs = src.crs
            else
               crs = self.crs
```

Then, just use `crs` (not `self.crs`) later.
```

File: src/stactools/core/utils/raster_footprint.py [hunk 2]
```diff
@@ -13,12 +13,277 @@
...
+
+            # create datamask
+            if self.no_data is None:
+                self.no_data = src.nodata
```
gadomski, line 152: Same as above, mutating `self` feel a little dangerous unless we need to.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+
+            if not src.indexes:
+                raise ValueError(
+                    "Raster footprint cannot be computed for an asset with no bands."
+                )
+            if not self.bands:
+                self.bands = src.indexes
```
gadomski, line 159: Same as above for `bands`.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+
+            if self.no_data is not None:
+                self.data: npt.NDArray[np.uint8] = np.full(src.shape, 0, dtype=np.uint8)
```
gadomski, line 162:
```suggestion
               data: npt.NDArray[np.uint8] = np.full(src.shape, 0, dtype=np.uint8)
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+        raise NotImplementedError
+
+    def simplify(self, polygon: Polygon) -> Polygon:
+        if self.simplify_tolerance is not None:
+            polygon = polygon.simplify(
+                tolerance=self.simplify_tolerance, preserve_topology=False
+            ).simplify(0)
+
+        # simplify does not remove duplicate sequential points, so do that
+        return Polygon([k for k, _ in groupby(polygon.exterior.coords)])
```
gadomski, line 206: So I know this feels like I'm all over the place w/ "free functions vs. class-based", but this method feels likt it could live as its own free function:

```python
def simplify(polygon: Polygon, simplify_tolerance: float) -> Polygon:
    ...
```

This would allow you to remove the `is not None` checks from the function (that could be performed in the class), and would provide the function to folks that just wanted to do simplification w/o all the other rigamarole.

I see below that you've created `_densify_by_factor` as a standalone, "private" function ... a similar `_simplify` function would make sense too, in other words.

src/stactools/core/projection.py
```diff
@@ -1,10 +1,21 @@
...
 
+# Sinusoidal projection parameters derived from Appendix 2, Section 13.1 in:
+# https://modis-fire.umd.edu/files/MODIS_Burned_Area_Collection51_User_Guide_3.1.0.pdf
+# All parameters specified to 12 significant digits so we can round to 11
+# significant digits after using them in computations. This assures millimeter
+# precision at the projection extremeties.
+SINUSOIDAL_SPHERE_RADIUS = 6371007.18100
+SINUSOIDAL_TILE_METERS = 1111950.51977
+SINUSOIDAL_X_MIN = -20015109.3558
+SINUSOIDAL_Y_MAX = 10007554.6779
```
gadomski, line 17: So now that I'm looking at it, I'm a _little_ hesitant to bring this functionality all the way up to **stactools**, as it feels very dataset-specific. I'm now leaning towards keeping this in **stactools-packages-modis**, and maybe **stactools-packages-viirs** depends on **modis** to get these values/functions?

gadomski (changes requested):

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+                    logger.error(
+                        f"Could not determine extent for asset '{name}', missing href"
+                    )
+                else:
+                    extent = self.data_footprint(href)
+                    if extent:
+                        yield name, extent
+                    else:
+                        logger.error(f"Could not determine extent for asset '{name}'")
+
```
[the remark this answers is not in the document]
pjhartzell: I like. Done.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+        polygon = self.simplify(polygon)
+
+        return mapping(polygon)  # type: ignore
+
+    def densify(self, polygon: Polygon) -> Polygon:
+        raise NotImplementedError
+
+    def reproject(self, polygon: Polygon) -> Polygon:
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
pjhartzell: Yep. Done.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+    ) -> None:
+        self.asset_names = asset_names
+        self.precision = precision
+        self.densification_factor = densification_factor
+        self.simplify_tolerance = simplify_tolerance
+        self.no_data = no_data
+        self.bands = bands
+        self.crs = crs
+
+    def update_geometry_from_asset_footprint(self, item: Item) -> bool:
```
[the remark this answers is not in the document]
pjhartzell: Done.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+    cum_interp_lengths = np.arange(0, total_length, densification_length)
+    cum_interp_lengths = np.append(cum_interp_lengths, [total_length])
+
+    interp_x = np.interp(cum_interp_lengths, cum_segment_lengths, points[:, 0])
+    interp_y = np.interp(cum_interp_lengths, cum_segment_lengths, points[:, 1])
+
+    return [(x, y) for x, y in zip(interp_x, interp_y)]
+
+
+class RasterFootprint:
```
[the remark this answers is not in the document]
pjhartzell: Done.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+                    extent = self.data_footprint(href)
+                    if extent:
+                        yield name, extent
+                    else:
+                        logger.error(f"Could not determine extent for asset '{name}'")
+
+    def data_footprint(self, href: str) -> Optional[Dict[str, Any]]:
+        with rasterio.open(href) as src:
+            self.transform = src.transform
+            self.shape = src.shape
```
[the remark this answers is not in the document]
pjhartzell: Given the large refactor, this comment may be moot.

src/stactools/core/utils/raster_footprint.py
pjhartzell: No longer exists in refactor.

src/stactools/core/utils/raster_footprint.py
pjhartzell: No longer an issue in the refactor.

src/stactools/core/utils/raster_footprint.py
pjhartzell: No longer an issue in the refactor.

src/stactools/core/utils/raster_footprint.py
pjhartzell: No longer an issue in the refactor.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -13,12 +13,277 @@
...
+                self.no_data = src.nodata
+
+            if not src.indexes:
+                raise ValueError(
+                    "Raster footprint cannot be computed for an asset with no bands."
+                )
+            if not self.bands:
+                self.bands = src.indexes
+
+            if self.no_data is not None:
```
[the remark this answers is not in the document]
pjhartzell: No longer an issue in the refactor.

src/stactools/core/projection.py
pjhartzell: Agreed. Removed.

src/stactools/core/utils/raster_footprint.py
pjhartzell: `simplify()` (now `simplify_polygon()`) is now a single shapely simplify call. I moved the duplicate point removal to where the duplicate points are generated, which is in the reprojection function (that's where the coordinates get rounded, which causes duplicate coords when coarsely rounding). For now I have _not_ extracted a free function from the contents of `reproject_polygon()`.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -19,76 +22,660 @@
...
+    represented by a polygon with two parallel sides and two curved sides. The
+    difference between these representations is greater the further away from
+    the meridian and equator the asset is located.
+
+    After reprojection to EPSG 4326, the footprint may have more points than
+    desired. This can be simplified to a polygon with fewer points that maintain
+    a maximum distance to the original geometry.
+
+    Args:
+        data_array (npt.NDArray[Any]): The raster data used for the
```
gadomski, line 115:
```suggestion
        data_array (numpy.NDArray[Any]): The raster data used for the
```

The documentation doesn't show `import _ as _`, so we need to provide the full Python module/package name.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -19,76 +22,660 @@
...
+                densify_by_factor(polygon.exterior.coords, self.densification_factor)
+            )
+        elif self.densification_distance is not None:
+            return Polygon(
+                densify_by_distance(
+                    polygon.exterior.coords, self.densification_distance
+                )
+            )
+        else:
+            return polygon
```
gadomski, line 287: Should we check to see if both are set, and error if that is the case? I think so...

src/stactools/core/utils/raster_footprint.py
```diff
@@ -19,76 +22,660 @@
...
+            polygon (Polygon): Footprint polygon in the native CRS.
+
+        Returns:
+            Polygon: Footprint polygon in EPSG 4326.
+        """
+        polygon = shape(
+            transform_geom(self.crs, "EPSG:4326", polygon, precision=self.precision)
+        )
+        # Rounding to precision can produce duplicate coordinates, so we remove
+        # them. Once once shapely>=2.0.0 is supported, this can be replaced with
```
gadomski, line 305:
```suggestion
        # them. Once once shapely>=2.0.0 is required, this can be replaced with
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -19,76 +22,660 @@
...
+        # shapely.constructive.remove_repeated_points
+        polygon = Polygon([k for k, _ in groupby(polygon.exterior.coords)])
+        return polygon
+
+    def simplify_polygon(self, polygon: Polygon) -> Polygon:
+        """Reduces the number of polygon vertices such that the simplified
+        polygon shape is no further away than the original polygon vertices than
+        the ``simplify_tolerance`` distance specified in the class constructor.
```
gadomski, line 313:
```suggestion
       ``self.simplify_tolerance``.
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -19,76 +22,660 @@
...
+        *,
+        no_data: Optional[Union[int, float]] = None,
+        bands: List[int] = [1],
+        precision: int = DEFAULT_PRECISION,
+        densification_factor: Optional[int] = None,
+        densification_distance: Optional[float] = None,
+        simplify_tolerance: Optional[float] = None,
+    ) -> "RasterFootprint":
+        """Produces a :py:class:`RasterFootprint` instance from an image href.
+
```
gadomski, line 340:
```suggestion
        """Produces a :py:class:`RasterFootprint` instance from an image href.

        The href can point to any file that is openable by rasterio.
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -278,38 +802,42 @@ def densify_reproject_simplify(
...
     if simplify_tolerance is not None:
         polygon = polygon.simplify(
             tolerance=simplify_tolerance, preserve_topology=False
-        ).simplify(0)
-
-    # simplify does not remove duplicate sequential points, so do that
-    polygon = Polygon([k for k, _ in groupby(polygon.exterior.coords)])
+        )
 
     return polygon
```
gadomski, line 843: Ideally, this code could share the same functionality as https://github.com/stac-utils/stactools/pull/396/files#diff-3ec600e1e4dab0a2b3d7aca0a4f8b7054c149fca25e23afbcc7d075c7e0b4e2dR210-R212 ... is that possible?

gadomski (changes requested):
Do you think we should deprecate the free functions now? Or wait until a later time? Also, looks like rasterio 1.3.5.post1 changed how some subdatasets are named: https://github.com/stac-utils/stactools/actions/runs/4156577655/jobs/7190456348#step:5:32? That feels strange to me...

src/stactools/core/utils/raster_footprint.py
pjhartzell: >Do you think we should deprecate the free functions now? Or wait until a later time?

I was going to hold off until the changes have been properly exercised via the MODIS stactools package.

>Also, looks like rasterio 1.3.5.post1 changed how some subdatasets are named: https://github.com/stac-utils/stactools/actions/runs/4156577655/jobs/7190456348#step:5:32? That feels strange to me...

Yeah, not sure how to handle this.

src/stactools/core/utils/raster_footprint.py
pjhartzell: This is done in `__init__()`

src/stactools/core/utils/raster_footprint.py
pjhartzell: Done

src/stactools/core/utils/raster_footprint.py
pjhartzell: Done.

src/stactools/core/utils/raster_footprint.py
pjhartzell: Not clear on your question here...

gadomski: The code in `RasterFootprint.reproject_polygon` (https://github.com/stac-utils/stactools/pull/396/files#diff-3ec600e1e4dab0a2b3d7aca0a4f8b7054c149fca25e23afbcc7d075c7e0b4e2dR303-R310) is the same, functionally, as the code in `densify_reproject_simplify` free function (https://github.com/stac-utils/stactools/pull/396/files#diff-3ec600e1e4dab0a2b3d7aca0a4f8b7054c149fca25e23afbcc7d075c7e0b4e2dR839-R840). Is there any way to only have that logic in one place?

pjhartzell: Ah. I see. I moved the common logic to a new free function.

src/stactools/core/utils/raster_footprint.py
pjhartzell: > Do you think we should deprecate the free functions now? Or wait until a later time?

Free functions that duplicate class methods are now marked as deprecated since version 0.4.4. Should we bump that to 0.5.0?

pjhartzell: >Also, looks like rasterio 1.3.5.post1 changed how some subdatasets are named

I modified the failing test to handle rasterio's old and new subdataset naming by stripping the leading underscores off the names (and sorting, but the new order doesn't bother me). The underscore removal in the test feels like a hack. If anything, we should look at stripping these leading underscores off in `cogify_subdatasets`, since we end up a[dding yet another underscore](https://github.com/stac-utils/stactools/blob/327985c01bfe739c01db29d3b72b4cc75ed8ef9a/src/stactools/core/utils/convert.py#L115) in the file name. 

Thoughts? Is the different underscoring a breaking change?

pjhartzell: >The documentation doesn't show import _ as _, so we need to provide the full Python module/package name.

This is fixed.

gadomski:
> > Free functions that duplicate class methods are now marked as deprecated since version 0.4.4. Should we bump that to 0.5.0?

We don't have an explicit deprecation policy, but IMO it's best to give at least one release cycle for the deprecations to kick in. So depreacted in 0.4.4, removal planned for 0.6.

> Thoughts? Is the different underscoring a breaking change?

I _think_ the problem was due to using **rasterio** wheels rather than building from scratch (f75b36c8d4d299ce32247dfe10827ba413876fdb). I agree that the test is a bit fragile but #401 _should_ fix it.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -18,6 +21,622 @@
...
+            polygon.
+        no_data (Optional[Union[int, float]]): The nodata value in
+            ``data_array``. If set to None, this will return a footprint
+            including nodata values.
+    """
+
+    crs: CRS
+    """Coordinate reference system of the raster data."""
+
+    data_array: npt.NDArray[Any]
```
gadomski, line 175: Were we not able to use `Scalar` here?

src/stactools/core/utils/raster_footprint.py
```diff
@@ -18,6 +21,622 @@
...
+                "can be specified."
+            )
+        self.densification_factor = densification_factor
+        self.densification_distance = densification_distance
+        self.simplify_tolerance = simplify_tolerance
+        self.no_data = no_data
+
+    def footprint(self) -> Optional[Dict[str, Any]]:
+        """Produces the footprint surrounding data (not nodata) pixels in
+        the source image. If the footprint was unable to be computed, None
```
gadomski, line 226:
```suggestion
        the source image. If the footprint is unable to be computed, None
```

src/stactools/core/utils/raster_footprint.py
```diff
@@ -18,6 +21,622 @@
...
+        """Adds vertices to the footprint polygon in the native CRS using either
+        ``self.densification_factor`` or ``self.densification_distance``.
+
+        Args:
+            polygon (Polygon): Footprint polygon in the native CRS.
+
+        Returns:
+            Polygon: Densified footprint polygon in the native CRS.
+        """
+        if self.densification_factor is not None:
```
gadomski, line 302:
```suggestion
        assert not (self.densification_factor and self.densification_distance)
        if self.densification_factor is not None:
```

This is a good place for an assertion. We do a check w/ a meaningful error in the constructor, but we still want to be sure we haven't set both -- `assert` is appropriate b/c it's cheap, and the only way we trigger it is if someone did something unexpected.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -18,6 +21,622 @@
...
+        no_data (Optional[Union[int, float]]): The nodata value in
+            ``data_array``. If set to None, this will return a footprint
+            including nodata values.
+    """
+
+    crs: CRS
+    """Coordinate reference system of the raster data."""
+
+    data_array: npt.NDArray[Any]
+    """3D array of raster data."""
```
gadomski, line 176: Could it be 2D? Or do 2D matrices need to be promoted to 3D?

src/stactools/core/utils/raster_footprint.py
```diff
@@ -18,6 +21,622 @@
...
+                one new point between each existing pair of points), a factor of
+                3 would place two points between each point, etc. Higher
+                densities produce higher fidelity footprints in areas of high
+                projection distortion.
+            simplify_tolerance (Optional[float]): Distance, in degrees, within which
+                all locations on the simplified polygon will be to the original
+                polygon.
+            no_data (Optional[Union[int, float]]): Explicitly sets the nodata value
+                if not in source image metadata. If set to None, this will return
+                a footprint including nodata values.
```
gadomski, line 524: This is missing `densification_distance`, which tells me that we're repeating these docs in too many places. I think it could be appropriate to enumerate the full meaning once (e.g. in the class definition) and just point to those docs with references.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -26,69 +645,63 @@ def update_geometry_from_asset_footprint(
...
-            item,
-            asset_names=asset_names,
-            precision=precision,
-            densification_factor=densification_factor,
-            simplify_tolerance=simplify_tolerance,
-            no_data=no_data,
-            bands=bands,
-        ),
-        None,
+    return RasterFootprint.update_geometry_from_asset_footprint(
```
gadomski, line 695: This should also use `warnings.warn` to programatically deprecate the function. Same is true for the other depreacated free functions.

tests/core/utils/test_convert.py
```diff
@@ -38,11 +38,12 @@ def test_subdataset(self):
...
                 paths, names = cogify_subdatasets(infile, directory)
+            names = [name.lstrip("_") for name in names]
             self.assertEqual(
-                names,
+                sorted(names),
                 [
-                    "__MonthlyRainTotal_GeoGrid_Data_Fields_RrLandRain",
-                    "__MonthlyRainTotal_GeoGrid_Data_Fields_TbOceanRain",
+                    "MonthlyRainTotal_GeoGrid_Data_Fields_RrLandRain",
+                    "MonthlyRainTotal_GeoGrid_Data_Fields_TbOceanRain",
```
gadomski, line 46: Per https://github.com/stac-utils/stactools/pull/396#issuecomment-1431713557 we can drop this fix, CI should let the original code pass now.

gadomski (changes requested):

src/stactools/core/utils/raster_footprint.py
pjhartzell: Done.

src/stactools/core/utils/raster_footprint.py
pjhartzell: Done.

src/stactools/core/utils/raster_footprint.py
pjhartzell: It can be 2D. When initialized, a 2D array will be promoted to 3D since `self.data_mask()` requires a 3D array for `np.sum()` call. I made this a bit more explicit:

- Changed the docstring to "2D or 3D array of raster data."
- Added an assert in `self.data_mask()` that the number of dimension is == 3.

src/stactools/core/utils/raster_footprint.py
pjhartzell: Good catch. I agree that it is a lot of repetition. Unfortunately the class definition does not contain all arguments, as some of them are only specified in the alternative constructors. And the alternative constructors themselves differ slightly in their arguments.

For now, I've pasted in the missing argument. I also like being able to see the full list of argument definitions via "intellisense", so I'd lean toward keeping the repetitive lists.

tests/core/utils/test_convert.py
pjhartzell: I removed my changes.

src/stactools/core/utils/raster_footprint.py
pjhartzell: Done.

CHANGELOG.md
```diff
@@ -320,22 +328,3 @@ See [#9](https://github.com/stac-utils/stactools/pull/9)
...
-[0.2.3]: <https://github.com/stac-utils/stactools/compare/v0.2.2..v0.2.3>
-[0.2.2]: <https://github.com/stac-utils/stactools/compare/v0.2.1..v0.2.2>
-[0.2.1]: <https://github.com/stac-utils/stactools/compare/v0.1.6..v0.2.1>
-[0.1.6]: <https://github.com/stac-utils/stactools/compare/v0.1.5..v0.1.6>
-[0.1.5]: <https://github.com/stac-utils/stactools/compare/v0.1.4..v0.1.5>
-[0.1.4]: <https://github.com/stac-utils/stactools/compare/v0.1.3..v0.1.4>
-[0.1.3]: <https://github.com/stac-utils/stactools/compare/v0.1.2..v0.1.3>
-[0.1.2]: <https://github.com/stac-utils/stactools/compare/v0.1.1..v0.1.2>
-[0.1.1]: <https://github.com/stac-utils/stactools/compare/v0.1.0..v0.1.1>
-[0.1.0]: <https://github.com/stac-utils/stactools/releases/tag/v0.1.0>
```
gadomski, line 341: Did you mean to remove all these links?

gadomski (changes requested):

gadomski (approved):

PR #415, opened 2023-04-06
Rework minimum requirements check

gadomski:
**Description:**
This simplifies things a bit (specifically, removes `requirements-min.txt` which felt a bit hackey). I explicitly don't handle non `>=` requirement specifications ... we'll cross that bridge when we get there?

I don't think these need a changelog.

**PR checklist:**


jpolchlo (changes requested):
While attempting to install into a clean virtual environment, there was a complaint that `packaging` was missing when I attempted to simply run `./scripts/install-min-requirements`.  Is this script is just intended for use in the CI?  I see that `requirements-dev.txt` includes `packaging`, and those requirements are pulled in ahead of running this script.  However, everything works via `pip install`.  I'm confused as to the function of this script, and if it is not meant for developers, should it live in `scripts`?  Also, the `scripts` directory was supposed to be consistent with [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all), and it might be better to stick to that convention by calling this script `bootstrap`.

This "request changes" response is provisional.  I'm happy to convert to an "approve" with just a few words of explanation.

gadomski:
Thanks for the review!

> While attempting to install into a clean virtual environment, there was a complaint that packaging was missing when I attempted to simply run ./scripts/install-min-requirements

I'll add a doc note to the top of script explaining that you need to install the dev dependencies first.

>  Is this script is just intended for use in the CI? 

Generally, yeah, though you could use it locally too if you want.

>  I'm confused as to the function of this script, and if it is not meant for developers, should it live in scripts?

Why not? It's a script :-)

> Also, the scripts directory was supposed to be consistent with [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all), and it might be better to stick to that convention by calling this script bootstrap.

That's not really documented anywhere in this repo, and we've had the practice of adding other scripts to `scripts/` in related projects (e.g. https://github.com/stactools-packages/noaa-cdr/blob/main/scripts/create_examples.py). It's not really a bootstrap script, either -- rather, it's intended as a tool to help ensure our library continues to work w/ the floored versions of its dependencies (which is sort of hard to do, e.g. https://github.com/pypa/pip/issues/8085). For context, I did a writeup of the original setup here: https://www.gadom.ski/2022/02/18/dependency-protection-with-python-and-github-actions.html.

jpolchlo (approved):

jpolchlo:
I wonder if you might include a link to this blog post as context as a comment in the min requirements script?  It's good context, and might be valuable to others who stumble in to class late (like me).  It clears up the function of what's being done here, and will probably dissuade others from naively attempting to use this script in a more general fashion.

PR #416, opened 2023-04-10
Add `enclose_poles` to antimeridian module

gadomski:
**Related Issue(s):**
- Required for https://github.com/microsoft/planetary-computer-tasks/pull/167

**Description:**
Items whose assets include data over the poles can be hard to represent in WGS84, which is required for their geometries. This PR adds a new function, `stactools.core.utils.antimeridian.enclose_poles`, that modifies geometries that cross the antimerdian to enclose the polar regions.

## Explanation

Here's some sentinel5p data:

image

As you can see, the data encloses both the north and the south poles. A simple geometry will follow the boundary of the data, but doesn't display (or really make sense at all) on a map:

Screenshot 2023-04-10 at 7 22 43 AM

To solve the problem, we can "box" the poles by detecting antimeridian crossings and adding points to cover the polar regions in WGS84:

Screenshot 2023-04-10 at 7 23 38 AM

This last image shows a geometry created with `enclose_poles`.

**PR checklist:**

### DIFF
--- a/tests/core/utils/test_antimeridian.py
+++ b/tests/core/utils/test_antimeridian.py
@@ -16,7 +16,7 @@ def test_antimeridian_split() -> None:
         (
             shapely.geometry.box(170, 40, 180, 50),
             shapely.geometry.box(-180, 40, -170, 50),
-        )
+        ),
     )
     for actual, expected in zip(split.geoms, expected.geoms):
         assert actual.exterior.is_ccw
@@ -27,7 +27,7 @@ def test_antimeridian_split() -> None:
     assert split is None
 
     canonical_other_way = Polygon(
-        ((-170, 40), (170, 40), (170, 50), (-170, 50), (-170, 40))
+        ((-170, 40), (170, 40), (170, 50), (-170, 50), (-170, 40)),
     )
     split = antimeridian.split(canonical_other_way)
     assert split
@@ -35,7 +35,7 @@ def test_antimeridian_split() -> None:
         (
             shapely.geometry.box(-180, 40, -170, 50),
             shapely.geometry.box(170, 40, 180, 50),
-        )
+        ),
     )
     for actual, expected in zip(split.geoms, expected.geoms):
         assert actual.exterior.is_ccw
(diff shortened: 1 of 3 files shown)


tests/core/utils/test_antimeridian.py
```diff
@@ -199,8 +199,30 @@ def test_antimeridian_multipolygon() -> None:
...
             Polygon(((170, 48), (170, 50), (190, 50), (190, 48), (170, 48))),
-        )
+        ),
     )
     for actual, expected in zip(normalized.geoms, expected.geoms):
         assert actual.exterior.is_ccw
         assert actual.equals(expected), f"actual={actual}, expected={expected}"
+
+
+def test_antimeridian_enclose_poles() -> None:
```
TomAugspurger, line 209: Could you add a test that ensures we hit both of the branches at https://github.com/stac-utils/stactools/pull/416/files#diff-39c4b9c63d34aa4ac3f9e5916273757b58071aeca592e256e8306f41729f0499R269, starting with

```
if start[0] > 0:
```
?

gadomski: We already hit both of those branches -- one of them is the north pole enclosure (starting at +170° longitude) and the other is the south pole enclosure (starting at -170° longitude). From a local coverage run:
Screenshot 2023-04-10 at 08-09-54 Coverage for src_stactools_core_utils_antimeridian py 97%

gadomski:
> Only other question would be whether we want a test that verifies we raise a ValueError when the polygon has an interior ring. I'm not 100% sure why that matters (and am OK remaining ignorant here), so I'll let you judge whether a test for that case is important.

I don't think its particularly necessary, as this is more of a one-off utility rather than a foundational API function. I'm sure there's plenty of other edge cases that would blow this function up that aren't tested.

That being said, I have a thought about one more edge case we _could_ test for -- the case where there's more than two antimeridian crossings (e.g. a swath over the Pacific). I'm going to convert to draft until I can cook up a test case (or better yet, a real-world example) of that and see if we can handle it.

gadomski:
@TomAugspurger I've added code and tests (and comments, because it's reasonably complex code) to handle multiple crossings. Ready for re-review, thanks.

tests/core/utils/test_antimeridian.py
TomAugspurger: Ah right, I forgot that a single polygon could cross both. Thanks!

TomAugspurger (approved):
Looks good, thanks!

PR #424, opened 2023-05-11
Use pyproject.toml

gadomski:
**Description:**
Includes a pre-commit autoupdate. No functional code changes.

**PR checklist:**

- [x] Code is formatted (run `scripts/format`).
- [x] Code lints properly (run `scripts/lint`).
- [x] Tests pass (run `scripts/test`).
- [x] Documentation has been updated to reflect changes, if applicable.
- [x] Changes are added to the [CHANGELOG](../CHANGELOG.md).

PR #426, opened 2023-05-12
Use the antimeridian package

### LINKED ISSUE #421: Use the antimeridian package
As a part of this month's Planetary Computer release, I developed a Python package for correcting complicated antimeridian-crossing and pole-crossing GeoJSON geometries: https://pypi.org/project/antimeridian/. It does almost everything our antimeridian fixer does, and more. We should update to use that package.

The only thing my package doesn't do is "normalization", aka going past 180 (or -180) instead of splitting. While "normalization" was desirable when we first made it, we've moved away from it, as it isn't technically spec-compliant. We could keep the normalization code around, or deprecate it for future removal.

gadomski:
**Related Issue(s):**
- Closes #421 
- Draft until #424 lands

**Description:**
We used to implement antimeridian fixes in this package, but we recently created https://github.com/gadomski/antimeridian. This package does more, is better tested, and has fewer dependencies (only shapely). This PR updates **stactools** to use that package.

The `fix_item` function is the most valuable -- the **antimeridian** package doesn't know anything about STAC, so **stactools** adds value by updating the item's geometry _and_ bounding box after the fix. The rest of the functions are deprecated, as folks should use **antimeridian** instead.

All "normalization" functions are deprecated, as "normalization" (using longitudes > 180 or <-180 to create a single polygon that crosses the antimeridian) doesn't conform to the GeoJSON spec.

**PR checklist:**


tests/core/utils/test_antimeridian.py
```diff
@@ -41,44 +45,10 @@ def test_antimeridian_split() -> None:
         assert actual.equals(expected), f"actual={actual}, expected={expected}"
 
 
-def test_antimeridian_split_complicated() -> None:
```
gadomski, line 44: These cases are tested in the **antimeridian** package, so I figured it wasn't worth updating them to be correct here.

tests/core/utils/test_antimeridian.py
```diff
@@ -8,8 +9,9 @@
 
 def test_antimeridian_split() -> None:
     # From https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.9
-    canonical = Polygon(((170, 40), (170, 50), (-170, 50), (-170, 40), (170, 40)))
-    split = antimeridian.split(canonical)
+    canonical = Polygon(((170, 40), (-170, 40), (-170, 50), (170, 50), (170, 40)))
```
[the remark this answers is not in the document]
ircwaves: 🇹🇭 Excellent!  Right-hand rule FTW!

tests/core/utils/test_antimeridian.py
ircwaves: Makes sense.

src/stactools/core/utils/antimeridian.py
```diff
@@ -248,104 +247,31 @@ def enclose_poles(polygon: Polygon) -> Polygon:
...
+            "To fix polygons that enclose the poles, use the antimeridian "
+            "package (https://pypi.org/project/antimeridian/)."
+        ),
+        DeprecationWarning,
+    )
+    fixed = antimeridian.fix_polygon(polygon)
+    if fixed.geom_type == "Polygon":
+        return fixed
+    else:
+        raise ValueError("Input polygon was split and we cannot return a MultiPolygon")
```
ircwaves, line 277: If this wasn't on the chopping block, I'd think maybe we could do a dirty `Union[Poly, Multi]`.

ircwaves (approved):
Deprecations are warm and fuzzy.

src/stactools/core/utils/antimeridian.py
gadomski: Agreed, and that's what I do here: https://github.com/gadomski/antimeridian/blob/931b78f83515b2037be428f7fe3db41d14f546db/src/antimeridian/_implementation.py#L215-L217. But yeah, since we're deprecating it didn't seem to make sense to change the return signature.

PR #440, opened 2023-06-23
Raster footprint: let user set destination CRS 4326

### LINKED ISSUE #435: Raster Footprint: Not EPSG4326
**Is your feature request related to a problem? Please describe.**
From [here](https://github.com/stac-utils/stactools/blob/b3d011e8b2a023e596b0ee269d7cb9a317b6c257/src/stactools/core/utils/raster_footprint.py#L113) it looks like the footprint is being forced to 4326 (called in line 234 of the same dir).

**Describe the solution you'd like**
If the CRS is being passed in, could we allow RasterFootprint to use the CRS for the reproject. The default should be "EPSG:4326" because the GeoJSON spec (unfortunately) assumed 4326. 

**Describe alternatives you've considered**
As an alternative to this, one could (I think?) take the output file and reproject is using something like `ogr2ogr`.

**Additional context**
None noted.

jsignell:

**Related Issue(s):**
Closes #435 

**Notes:**
- The output of the footprint method is still geojson, it's just invalid geojson if the user has declared a different dest crs. I am not sure if that is the right approach or not.
- I didn't update the deprecated functions. Don't see a reason to enable this enhancement on deprecated functions.
- I wrote a test, but I'd appreciate an extra set of eyes to make sure the expectation is sane. Or maybe there is a better way to test.

**PR checklist:**


gadomski (approved):
> The output of the footprint method is still geojson, it's just invalid geojson if the user has declared a different dest crs. I am not sure if that is the right approach or not.

I think this is just fine, since we're not outputting a "GeoJSON" structure or anything -- it's just a `Polygon`. The docs seem correct as well, they don't say "it's valid GeoJSON the way out."

> I didn't update the deprecated functions. Don't see a reason to enable this enhancement on deprecated functions.

I agree with this approach.

> I wrote a test, but I'd appreciate an extra set of eyes to make sure the expectation is sane. Or maybe there is a better way to test.

What you did feels right to me.

PR #443, opened 2023-06-27
Implement migrate command

### LINKED ISSUE #319: Implement or remove the migrate CLI command
Right now `stac migrate` is unimplemented: https://github.com/stac-utils/stactools/blob/a885f7636a822d6a02fba4ec40af4852e6570e6c/src/stactools/cli/commands/migrate.py#L4-L6. We should either implement or remove this functionality.

I think implementation would be relatively simple, at least in the base-case:
1. Open a catalog with PySTAC
2. Re-write it to the same location with PySTAC

We would just rely on PySTAC to handle all the conversions for us. We would want to provide some test cases that demonstrate how it works -- example old STAC catalogs could be gleaned from the spec.

jsignell:
**Related Issue(s):**
Closes #319 

**Description:**
Looks like

```python
stac migrate tests/data-files/planet-disaster/collection.json
```

<details><summary>Output</summary>

The red/green is coming from github markdown :)

```diff
--- before

+++ after

@@ -4,6 +4,11 @@

   "stac_version": "1.0.0",
   "description": "[Planet Disaster Data](https://www.planet.com/disasterdata/) makes imagery available directly to the public, volunteers, humanitarian organizations, and other coordinating bodies in support of the International Charter for Space and Major Disasters. Data is released for individual disaster events, providing a 30 day window pre- and post-disaster. Imagery is provided under Creative Commons licenses, free of charge, with either CC-BY-SA or CC-BY-NC.",
   "links": [
+    {
+      "rel": "self",
+      "href": "/home/jsignell/stactools/tests/data-files/planet-disaster/collection.json",
+      "type": "application/json"
+    },
     {
       "rel": "root",
       "href": "./collection.json",
@@ -16,8 +21,28 @@

       "title": "Hurricane Harvey"
     }
   ],
-  "stac_extensions": [],
   "title": "Planet Disaster Data",
+  "extent": {
+    "spatial": {
+      "bbox": [
+        [
+          -96,
+          28,
+          -93,
+          31
+        ]
+      ]
+    },
+    "temporal": {
+      "interval": [
+        [
+          "2017-08-28T18:00:00Z",
+          null
+        ]
+      ]
+    }
+  },
+  "license": "CC-BY-SA-4.0",
   "keywords": [
     "disaster",
     "open"
@@ -79,26 +104,5 @@

       "minimum": 0,
       "maximum": 24
     }
-  },
-  "extent": {
-    "spatial": {
-      "bbox": [
-        [
-          -96,
-          28,
-          -93,
-          31
-        ]
-      ]
-    },
-    "temporal": {
-      "interval": [
-        [
-          "2017-08-28T18:00:00Z",
-          null
-        ]
-      ]
-    }
-  },
-  "license": "CC-BY-SA-4.0"
+  }
 }
```

</details>

To save the file and not show the diff:

```python
stac migrate tests/data-files/planet-disaster/collection.json --save --hide-diff
```

**PR checklist:**


jsignell:
I kind of feel that we do not need to test the migrate behavior itself in this repo since that should be well tested in pystac itself.

jsignell:
Oh also let me know what you want the default behavior to be.

src/stactools/core/migrate.py
```diff
@@ -0,0 +1,37 @@
...
+        show_diff : Whether to print the diff between the original and new STAC
+            object. Defaults to True.
+    Returns:
+        STACObject : The migrated object - modified inplace.
+    """
+    # migration happens implicitly on load
+    stac_object = pystac.read_file(href)
+    if show_diff:
+        input = DefaultStacIO().read_text(href)
+        output = json.dumps(stac_object.to_dict(), indent=2)
```
gadomski, line 30: Do we have to worry about key order here? I think if we're diffing, we probably have to sort keys?

src/stactools/cli/commands/migrate.py
```diff
@@ -1,15 +1,25 @@
...
-    @cli.command("migrate", short_help="Migrate a STAC catalog (TODO)")
-    @click.option("--option", "-o", default=1, help="A test option")
-    def migrate_command(option: click.Option) -> None:
-        migrate(option)
+    @cli.command("migrate", short_help="Migrate a STAC object")
+    @click.argument("href")
+    @click.option(
+        "-s",
+        "--save",
+        is_flag=True,
```
gadomski, line 11: This make sense to me as a default.

src/stactools/cli/commands/migrate.py
```diff
@@ -1,15 +1,25 @@
...
+        help="Save migrated STAC object in original location.",
+    )
+    @click.option(
+        "--show-diff/--hide-diff",
+        default=True,
```
gadomski, line 16: This also makes sense to me as a default.

src/stactools/cli/commands/migrate.py
```diff
@@ -1,15 +1,25 @@
...
+        help=(
+            "Whether to dump diff between original and migrated object to stdout. "
+            "Defaults to --show-diff. "
+        ),
+    )
+    def migrate_command(href: str, save: bool, show_diff: bool) -> None:
```
gadomski, line 22: Probably want to add a `recursive` option to migrate a whole catalog.

gadomski (changes requested):

src/stactools/core/migrate.py
jsignell: Well... yeah I put a note to try to make it clear that key order differences will result in a diff. I am not sure we should alter the order just for the diff though because that could mean that the diff is empty but there is a change if you turn on save. So I wasn't sure what to do. I have been treating the dict order that pystac outputs as the "best-practice" order.

src/stactools/cli/commands/migrate.py
jsignell: yikes. Ok yeah sorry I thought that came for free. More tests needed...

tests/data-files/planet-disaster-v1.0.0-beta.2/collection.json
jsignell, line 1: This is from pystac data-files

src/stactools/cli/commands/migrate.py
```diff
@@ -1,15 +1,48 @@
...
+        default=True,
+        help=(
+            "Whether to dump diff between original and migrated object to stdout. "
+            "Defaults to --show-diff. "
+        ),
+    )
+    def migrate_command(
+        href: str, save: bool, recursive: bool, show_diff: bool
+    ) -> None:
+        _migrate(href, save=save, recursive=recursive, show_diff=show_diff)
```
gadomski, line 46: If `save=False` and `show_diff=False`, is this a no-op other than doing a lot of work and then finishing silently?

jsignell: yes. I will change it so that if both are false it raises instead.

src/stactools/cli/commands/migrate.py
```diff
@@ -1,15 +1,54 @@
...
+        stac_object, save=save, recursive=recursive, show_diff=show_diff
+    )
 
 
 def create_migrate_command(cli: click.Group) -> click.Command:
-    @cli.command("migrate", short_help="Migrate a STAC catalog (TODO)")
-    @click.option("--option", "-o", default=1, help="A test option")
-    def migrate_command(option: click.Option) -> None:
-        migrate(option)
+    @cli.command("migrate", short_help="Migrate a STAC object")
```
gadomski, line 27:
```suggestion
    @cli.command("migrate", short_help="Migrate a STAC object to the latest version")
```

gadomski (approved):

Issue #392, opened 2023-01-26
Raster footprint feature: Union multiple asset geometries


pjhartzell:
**Is your feature request related to a problem? Please describe.**
I want to generate a single raster footprint polygon union'd from polygons generated from multiple assets in a single Item. For example, an Item may have several reflectance assets for differing wavelengths, each of which may have different areas of valid data. I'd like my raster footprint to encompass the footprint of all valid data for all the reflectance assets.

The current version of the `update_geometry_from_asset_footprint()` accepts a list of assets to operate on via the `asset_names` parameter, but returns after the first successful geometry creation.

**Describe the solution you'd like**
An option in the `update_geometry_from_asset_footprint()` signature to compute geometries for all assets in the `asset_names` list parameter and return their union.

**Describe alternatives you've considered**
Roll your own merge ([example](https://github.com/stactools-packages/aster/blob/66dfa283b2ad426280315f9b65da89b20fc87eda/src/stactools/aster/utils.py#L88-L129)). But the desired feature seems like a straightforward addition to the function.

PR #445, opened 2023-06-27
Allow geometries to be computed from multiple asset footprints

jpolchlo:
**Related Issue(s):**
#392 

**Description:**
This PR extends `update_geometry_from_asset_footprint()` to have more flexibility in how the footprint is derived in the event of multiple, non-coincident asset footprints.  Previously, this method would simply take the first matching asset footprint.  This PR allows for the footprint to be the intersection or union of the matching assets.  Adding the intersection was an executive decision, since I can imagine wanting to build STACs for analysis where we want to be sure data is present for all assets to avoid needing to filter for NODATA in subsequent steps.  If this doesn't make sense, it can be removed in favor of only using the union.

**PR checklist:**


jpolchlo:
This PR needs an improved test.  I'm in the market for a good sample Item to base it on.

src/stactools/core/utils/raster_footprint.py
```diff
@@ -567,32 +570,47 @@ def update_geometry_from_asset_footprint(
...
+                return False
+            _, extent = asset_name_extent
         else:
-            return False
+            extents = [shape(extent) for _, extent in asset_extent_iterator]
+            if extents == []:
+                return False
+            if footprint_aggregator == "intersection":
+                extent = mutual_intersection(extents)
+            elif footprint_aggregator == "union":
```
gadomski, line 605: Should be an enum.

gadomski (changes requested):
As mentioned, needs a test case.

gadomski (approved):

jpolchlo:
I realized too late that the baseline for the test here should have been derived from rasters that indicate the union/intersection.  Not worth holding up the show, though.  Merging this, but making this comment for posterity.

PR #455, opened 2023-08-10
Don't use deprecated methods from pystac

gadomski:
**Description:**
~Disallow warnings in tests, which encourages (nay, forces) us to stop using deprecated items.~ There are some warnings from **stac-check** that I'm having trouble filtering out, so for now I'm dropping the "upgrade to error" bit and just fixing the deprecated pystac function usages.

**PR checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `reproject_shape` without a precision ([#454](https://github.com/stac-utils/stactools/pull/454))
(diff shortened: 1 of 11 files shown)


jsignell (approved):
Nice! I think you can set the behavior of warnings based on what library they come from. So for instance you could set only pystac warnings to error. It looks like this: https://github.com/dask/dask/blob/ca4d1d5830a2efe9650aa6485302193c2aacaf7b/pyproject.toml#L129

gadomski:
Cool, I'll try the glob. I was trying to do it with exact name and failing.

gadomski:
Tight, that worked! I tested by cherry picking 6b17e0575f165d1bd4c6a56b70482b03b8c4b4bd onto **main**. Thanks @jsignell!

PR #468, opened 2023-11-15
Make stac-validator and stac-check optional

gadomski:
**Description:**
They have some unnecessary required dependencies that pollute downstreams. cc @pjhartzell @jkeifer 

**PR checklist:**

### DIFF
--- a/src/stactools/cli/__init__.py
+++ b/src/stactools/cli/__init__.py
@@ -2,6 +2,21 @@
 
 import stactools.core
 
+try:
+    import stac_validator as _
+except ImportError:
+    HAS_STAC_VALIDATOR = False
+else:
+    HAS_STAC_VALIDATOR = True
+
+try:
+    import stac_check as _
+except ImportError:
+    HAS_STAC_CHECK = False
+else:
+    HAS_STAC_CHECK = True
+
+
 stactools.core.use_fsspec()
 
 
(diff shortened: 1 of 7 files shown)


jsignell (approved):
LGTM

src/stactools/cli/__init__.py
```diff
@@ -35,15 +48,23 @@ def register_plugin(registry: "Registry") -> None:
...
     registry.register_subcommand(summary.create_summary_command)
-    registry.register_subcommand(validate.create_validate_command)
     registry.register_subcommand(version.create_version_command)
     registry.register_subcommand(update_extent.create_update_extent_command)
     registry.register_subcommand(update_geometry.create_update_geometry_command)
 
+    if HAS_STAC_VALIDATOR:
+        from stactools.cli.commands import validate
+
+        registry.register_subcommand(validate.create_validate_command)
```
jsignell, line 61: I guess my only question is what happens if a user tries to run `stac validate` and they don't have the deps? Ideally that would result in a nice error message that tells them what to do. So maybe that would lend itself to always registering and then have the error raised within the actual call to do validation?

gadomski: Hm, that's a good point. I can make a follow-on that makes the UX a bit nicer.

jsignell: I see people writing `import_optional` helpers a lot. Ala https://github.com/stac-utils/xpystac/blob/main/xpystac/utils.py#L5-L10

